PR #6, opened 2024-02-24
Update configs and scripts

micahdbak:
I've gone through the scripts used for this server as well as the configuration files to remove lines that are generated automatically, as well as redundant parts of scripts.

I've verified that the current set of configs and scripts works on a fresh Debian 12 machine on DigitalOcean, and deployed the backend to https://api2.micahdb.com.

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -15,8 +15,8 @@ server {
     include /etc/letsencrypt/options-ssl-nginx.conf;
     ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;
     
-    access_log /home/csss-site/csss-site-backend/logs/nginx-access.log;
-    error_log /home/csss-site/csss-site-backend/logs/nginx-error.log;
+    access_log /var/www/logs/csss-site-backend/nginx-access.log;
+    error_log /var/www/logs/csss-site-backend/nginx-error.log;
 
     location / {
         proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -9,12 +9,6 @@ server {
     keepalive_timeout 5;
     client_max_body_size 1G; # Was 4G
 
-    # managed by Certbot
-    ssl_certificate /etc/letsencrypt/live/api.sfucsss.org/fullchain.pem;
-    ssl_certificate_key /etc/letsencrypt/live/api.sfucsss.org/privkey.pem;
-    include /etc/letsencrypt/options-ssl-nginx.conf;
-    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem;
-    
     access_log /var/www/logs/csss-site-backend/nginx-access.log;
     error_log /var/www/logs/csss-site-backend/nginx-error.log;
 
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -4,7 +4,7 @@ upstream backend {
 
 server {
     server_name api.sfucsss.org;
-    listen 443 ssl;
+    listen 80;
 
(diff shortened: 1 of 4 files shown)


fresh_setup.sh
```diff
@@ -1,96 +1,110 @@
...
+sudo -u csss-site git clone https://github.com/CSSS/csss-site-backend.git csss-site-backend
+
+echo "----"
+echo "configure sudo for csss-site..."
+cp csss-site-backend/config/sudoers.conf /etc/sudoers.d/csss-site
+
+echo "----"
+echo "configure nginx..."
+cp csss-site-backend/config/nginx.conf /etc/nginx/sites-available/csss-site-backend
+echo "You'll need to fill out the certbot configuration manually."
```
EarthenSky, line 63: could you link to documents describing how the config should be filled out?

micahdbak: 100%! Also, I'm working on this part right now, as certbot doesn't configure our config. It configures /etc/nginx/sites-enabled/default. So I'm going to try and play with our config and remove default so it actually pastes information into ours.

config/nginx.conf
```diff
@@ -9,14 +9,8 @@ server {
     keepalive_timeout 5;
     client_max_body_size 1G; # Was 4G
 
-    # managed by Certbot
-    ssl_certificate /etc/letsencrypt/live/api.sfucsss.org/fullchain.pem;
```
EarthenSky, line 13: I'm assuming these lines got moved somewhat that is managed by certbot automatically, could you tell me a bit more about how the new certbot setup will work?

fresh_setup.sh
EarthenSky: ah! I see. Do be careful that anything automatic isn't checked into source code. Or if it is, certbot knows not to overwrite it.

config/nginx.conf
micahdbak: Yes; now, the preferred server configuration is listed under an HTTP server (`listen 80;`), which is nginx-valid (it doesn't use SSL), but that certbot adapts to *become* an HTTPS server (`listen 443;`). This keeps the configs on the repo super simple and leaves much of the SSL configuration to how certbot handles it :)

config/update_config.sh
```diff
@@ -1,12 +1,28 @@
 #!/bin/bash
 
+# make sure user is root
```
EarthenSky, line 3: if we ever set up an automated system for deploying changes, we'll want to be careful with automating any config changes. I like the idea of requiring root. Perhaps all config changes should always be done manually

config/update_config.sh
```diff
@@ -1,12 +1,28 @@
...
-sudo systemctl restart csss-site
+systemd-analyze verify /home/csss-site/csss-site-backend/config/csss-site.service
+# only use new service if it is valid
+if [ $? -eq 0 ]; then
+	cp /home/csss-site/csss-site-backend/config/csss-site.service /etc/systemd/system/csss-site.service
+	systemctl restart csss-site
+fi
 
+echo "3. update sudo config"
+cp /home/csss-site/csss-site-backend/config/sudoers.conf /etc/sudoers.d/csss-site
```
EarthenSky, line 28: \<3

config/update_config.sh
micahdbak: agree with this!

config/update_config.sh
```diff
@@ -1,12 +1,28 @@
 #!/bin/bash
 
+# make sure user is root
+user=$(whoami)
```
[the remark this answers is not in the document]
micahdbak: When the configs in the repo differ from the configs on the server. Most of these are only modifyable by the root, hence the check that the root user is running the script.

PR #29, opened 2024-06-16
Expanding /auth/check route

micahdbak:
Test plan:
- [x] Deployed locally works
- [x] Passes branch + PR GH Actions checks
- [x] Works with [this PR to csss-raspberry](https://github.com/CSSS/csss-raspberry/pull/18)

### DIFF
--- a/src/auth/auth.py
+++ b/src/auth/auth.py
@@ -80,7 +80,7 @@ async def check_authentication(
 
     if session_id:
         await crud.task_clean_expired_user_sessions(db_session)
-        response_dict = await crud.check_session_validity(db_session, session_id)
+        response_dict = await crud.check_user_validity(db_session, session_id)
     else:
         response_dict = {"is_valid": False}
 
--- a/src/auth/auth.py
+++ b/src/auth/auth.py
@@ -80,7 +80,7 @@ async def check_authentication(
 
     if session_id:
         await crud.task_clean_expired_user_sessions(db_session)
-        response_dict = await crud.check_user_validity(db_session, session_id)
+        response_dict = await crud.check_user_session(db_session, session_id)
     else:
         response_dict = {"is_valid": False}
 
--- a/src/auth/crud.py
+++ b/src/auth/crud.py
@@ -73,8 +73,8 @@ async def check_user_session(db_session: AsyncSession, session_id: str) -> dict:
         return {
             "is_valid": True,
             "computing_id": existing_user_session.computing_id,
-            "first_logged_in": existing_user.first_logged_in,
-            "last_logged_in": existing_user.last_logged_in
+            "first_logged_in": existing_user.first_logged_in.isoformat(),
+            "last_logged_in": existing_user.last_logged_in.isoformat()
         }
     else:
         return {"is_valid": False}
--- a/src/auth/crud.py
+++ b/src/auth/crud.py
@@ -2,6 +2,7 @@ from datetime import datetime, timedelta
 
 from auth import models
 
+import logging
 import sqlalchemy
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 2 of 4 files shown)


src/auth/auth.py
```diff
@@ -80,7 +80,7 @@ async def check_authentication(
 
     if session_id:
         await crud.task_clean_expired_user_sessions(db_session)
-        response_dict = await crud.check_session_validity(db_session, session_id)
+        response_dict = await crud.check_user_validity(db_session, session_id)
```
EarthenSky, line 83: Should this function be `crud.check_user_session`?

src/auth/crud.py
```diff
@@ -41,12 +51,19 @@ async def remove_user_session(db_session: AsyncSession, session_id: str) -> dict
...
 
 
-async def check_session_validity(db_session: AsyncSession, session_id: str) -> dict:
+async def check_user_session(db_session: AsyncSession, session_id: str) -> dict:
     query = sqlalchemy.select(models.UserSession).where(models.UserSession.session_id == session_id)
     existing_user_session = (await db_session.scalars(query)).first()
 
     if existing_user_session:
-        return {"is_valid": True, "computing_id": existing_user_session.computing_id}
+        query = sqlalchemy.select(models.User).where(models.User.computing_id == existing_user_session.computing_id)
```
EarthenSky, line 59: instead of explicitly doing another select, we can replace this with an sqlachemy relationship access. Let's leave it this way for now, but include a TODO.

src/auth/crud.py
```diff
@@ -17,6 +17,14 @@ async def create_user_session(db_session: AsyncSession, session_id: str, computi
     if existing_user_session:
         existing_user_session.issue_time = datetime.now()
         existing_user_session.session_id = session_id
+        query = sqlalchemy.select(models.User).where(models.User.computing_id == computing_id)
+        existing_user = (await db_session.scalars(query)).first()
+        if existing_user is None:
+            # @geb; throw?
```
EarthenSky, line 23: let's create a user row for them (like line 39), but log a warning for this case (since it should never occur unless something went wrong) using
```py
import logging
...
_logger = logging.getLogger(__name__)
...
# something like this
_logger.warning("Something is broken, aaaaaaaaaa")
```

`__name__` is the name of the current python module, so we can track it down more easily.

EarthenSky (changes requested):
Just some small changes. tysm for getting this PR done & learning part of the backend codebase!

src/auth/auth.py
micahdbak: yes! good catch

src/auth/crud.py
micahdbak: ooo sure, I'll drop a TODO

src/auth/crud.py
micahdbak: cool done!

src/auth/crud.py
```diff
@@ -20,7 +21,16 @@ async def create_user_session(db_session: AsyncSession, session_id: str, computi
         query = sqlalchemy.select(models.User).where(models.User.computing_id == computing_id)
         existing_user = (await db_session.scalars(query)).first()
         if existing_user is None:
-            # @geb; throw?
+            # log this strange case
+            _logger = logging.getLogger(__name__)
+            _logger.warning("User session {} exists for non-existent user {}!".format(session_id, computing_id))
```
EarthenSky, line 26: This is not important, but python now has f strings, like `f"User session {session_id} exists for non-existent user {computing_id}!"`, which is pretty nifty

micahdbak: Oooo that is nifty!

micahdbak: Lemme do that real quick for u

EarthenSky (approved):
looks good! for the time being we have no tests, so make sure it runs on your device haha

PR #30, opened 2024-06-20
Configure Nginx CORS

micahdbak:
I'm going to manually deploy these changes to nginx as I'm doing some testing with the CSSS auth API, but this should, in theory, make configuring CORS for the CSSS API super easy.

In the future we should probably replace '*' with just sfucsss.org, but for now * is good for testing

Update:

Local devs won't even be able to get CORS to work with localhost (i think cors policy needs HTTPS / an actual hostname that isnt localhost to send cookie credentials) so I went ahead and restricted it to new.sfucsss.org. Currently deployed & working :)

### DIFF
--- a/config/nginx.conf
+++ b/config/nginx.conf
@@ -19,9 +19,10 @@ server {
             break;
         }
 
-	# allow the CSSS API to be accessed from any origin
 	proxy_hide_header Access-Control-Allow-Origin;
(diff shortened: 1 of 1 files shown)


micahdbak:
Can confirm, this works with api.sfucsss.org

Screenshot 2024-06-20 at 1 40 15 PM

EarthenSky (approved):

Issue #31, opened 2024-07-04
interface with CSSS-WEBSITE discord application


EarthenSky:
CSSS discord has several bots (called discord applications): Wall-e, eve, and csss-website. Wall-e is a whole project undergoing renovations, and with different goals than the website. 

We'd like to query the csss discord server with things like "which people have which executive roles", or "who has access to which channels". We'll want to display this information to the dashboard so that executives can be notified if someone has permission that they shouldn't (ie. transitioned officer). Historically this was automated, however the execs / mods would prefer to have direct control. In addition, we'll want some api calls to modify permissions as well (a dashboard interface), but we should leave that for a future issue.

It looks like [we can](https://github.com/CSSS/csss-site/blob/189f0d2501d0cdefc15adbb5f06fd41dae65cddd/csss-site/src/about/views/input_new_officers/specify_new_officers/validators/validate_discord_id.py#L20) use [discord's http api](https://discord.com/developers/docs/quick-start/overview-of-apps#what-apis-can-apps-use) to get the info we're looking for.

Requirements:
- create a module which gets info from the csss discord guild. We can call this the "discord module"
  - we want the following functions:
    - get all discord channels in a category
    - get all users with access to a specific channel
    - ? get all users in our discord guild (server)
    - ? get all roles a user has in our discord guild
    - get all users who have a specific role
    - get all roles in our discord guild
- ~~figure out a way to mock this module so we can locally test it without making API calls (unless desired)~~

If you're not sure what we're looking for specifically, please ask so we're all on the same page. If you have any questions please lmk!

EarthenSky:
Looks like we might not even need a discord bot, as we'll use a token to access the discord api https://github.com/CSSS/csss-site/blob/189f0d2501d0cdefc15adbb5f06fd41dae65cddd/csss-site/src/about/views/input_new_officers/specify_new_officers/validators/validate_discord_id.py#L20

EarthenSky:
We need to stay beneath the rate limit of 50 requests per second (& 10k per 10 minutes), which should (hopefully) be pretty easy https://discord.com/developers/docs/topics/rate-limits#rate-limits

DerpyWasHere:
Specifically, this is just asking for the api to return data right? Any specific format you want that back in? Otherwise I'll default to JSON.

micahdbak:
^^^ yo good question, @EarthenSky are you asking for this to be an API or functions to be used internally by the server?

EarthenSky:
[...]

PR #41, opened 2024-07-14
Discord User Search / Refactoring of Discord Module

### LINKED ISSUE #38: Get discord id & server nickname given a username.
We want new executives to input their discord username, then for us to figure out their nickname in the CSSS Discord (or if they're not in the CSSS discord) & their discord id.

If the user is not in the CSSS Discord, we'll ask them to join first.

This is related to officer automation, so it should have a (slightly) higher priority.

DerpyWasHere:
Closed #38 and #31

I did a lot of refactoring of the discord module. A lot of it included generating custom objects for Channel, GuildMember, and User. Then refactoring functions for standardization and readability.

This depends on #40.


EarthenSky:
Protip: if we change the target branch, the git diff will only show the changes & commits you just made. Very convenient! Though we'll have to update it to `dev` before merging so we don't target the wrong location.

src/disc/disc.py
```diff
@@ -11,6 +11,44 @@
...
+        self.type = type
+        self.name = name
+        self.permission_overwrites = permission_overwrites
+
+    def __str__self(self) -> str:
+        return f"{self.id}, {self.name}"
+    
+    def __repr__(self) -> str:
+        return f"{self.id}, {self.name}"
+
```
EarthenSky, line 51: Let's update these to dataclasses, so we can be explicit about typing! https://docs.python.org/3.11/library/dataclasses.html

src/disc/disc.py
```diff
@@ -11,6 +11,44 @@
...
+        return f"{self.user.username}"
+    
+class Channel:
+    def __init__(self, id: str, type: int, guild_id: str, name: str, permission_overwrites: list[str] = None) -> None:
+        self.id = id
+        self.type = type
+        self.name = name
+        self.permission_overwrites = permission_overwrites
+
+    def __str__self(self) -> str:
```
EarthenSky, line 46: typo! make sure that you test every function manually (or by making unit tests) before merging \:)

src/disc/disc.py
```diff
@@ -11,6 +11,44 @@
...
+        self.username = username
+        self.discriminator = discriminator
+        self.global_name = global_name
+        self.avatar = avatar
+    
+    def __str__(self) -> str:
+        return f"{self.username}, {self.id}"
+    
+    def __repr__(self) -> str:
+        return f"{self.username}, {self.id}"
```
EarthenSky, line 26: Instead of overriding the defaults, should we make a custom function (say `def compact_str(): ...`) so that any debug printing will show all the data? I'm not sure it would make sense.

src/disc/disc.py
```diff
@@ -11,6 +11,44 @@
 ADMINISTRATOR = 1 << 3
 VIEW_CHANNEL = 1 << 10
 
+class User:
+    def __init__(self, id: str, username: str, discriminator: str, global_name: str=None, avatar: str=None) -> None:
+        self.id = id
+        self.username = username
+        self.discriminator = discriminator
```
EarthenSky, line 18: what's a discriminator? I wasn't able to figure it out by looking at where `discriminator` is used, so let's add a short comment.

src/disc/disc.py
```diff
@@ -116,18 +154,18 @@ async def get_role_name_by_id(
     id: str = guild_id
 ) -> str:
     roles = await get_all_roles(id)
-    return roles[rid]
+    return roles[rid][0]
```
EarthenSky, line 157: why are we adding a `[0]` when `get_all_roles` hasn't changed?

EarthenSky (changes requested):
Just some preliminary requests. I notice that you made a few changes here that I requested in the last pr #40.

I would suggest changing them in the old PR, then rebasing this over the old one, so that your commits where this happens don't exist anymore. However, that's a lot of work if you don't know how to use git very well, so we can just split them up if you'd prefer.

src/disc/disc.py
DerpyWasHere: Discriminators were deprecated, they used to be the #xxxx after a username. Accounts which haven't migrated over yet have them still.

EarthenSky: ahhh, I didn't know those were called discriminators, cool! Def add a lil comment. Glad we're handling them!

src/disc/disc.py
DerpyWasHere: Fixed :sob:

src/disc/disc.py
DerpyWasHere: After updating to use dataclasses, the autogenerated `__repr__` does this for us.

src/disc/disc.py
DerpyWasHere: Done.

src/disc/disc.py
DerpyWasHere: This returns a string instead of the list containing the name and the permissions and aligns with the return type described.

src/disc/disc.py
```diff
@@ -3,15 +3,36 @@
...
+ADMINISTRATOR = 0b1000
+VIEW_CHANNEL = 0b0010_0000_0000
+
+@dataclass
+class User:
+    id: str
+    username: str
+    discriminator: str
+    global_name: str = None
+    avatar: str = None
```
EarthenSky, line 20: In python `|` refers to the Union type. Since the default for `global_name` is None, the actual type is either None or str.

```suggestion
    global_name: str | None = None
    avatar: str | None = None
```

EarthenSky (changes requested):

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+            'Authorization': f'Bot {tok}',
+            'Content-Type': 'application/json',
+            'User-Agent' : 'DiscordBot (https://github.com/CSSS/csss-site-backend, 1.0)'
+        }
+    )
+    return result
+
+async def get_channel_members(
+    cid: str,
+    id: str = guild_id
```
EarthenSky, line 51:
```suggestion
    # TODO: hardcode guild_id (remove it as argument) if we ever refactor this module
    id: str = guild_id
```

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+    while True:
+        url = f'https://discord.com/api/v10/guilds/{id}/members?limit=1000&after={last_uid}'
+        result = await _discord_request(url, tok)
+
+        json_s = result.json()
+        res = [GuildMember(User(user['user']['id'], user['user']['username'], user['user']['discriminator'], user['user']['global_name'], user['user']['avatar']), user['roles'])
+                    for user in json_s if rid in user['roles']] 
+        matched += res
+
+        if res == []:
```
EarthenSky, line 200: can we move this check before `res = ...` & make it `if len(json_s) == 0`, to be more explicit? (slightly more performant, but that's not super important)

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+
+    json_s = result.json()
+    roles = [([role['id'], [role['name'], role['permissions']]]) for role in json_s]
+    return dict(roles)
+
+async def get_guild_members_with_role(
+    rid: str,
+    id: str = guild_id
+) -> list[GuildMember]:
+    tok = os.environ.get('TOKEN')
```
EarthenSky, line 179: Can we change `tok` to `token`?

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+
+async def get_role_by_id(
+    rid: str,
+    id: str = guild_id
+) -> dict:
+    tok = os.environ.get('TOKEN')
+    url = f'https://discord.com/api/v10/guilds/{id}/roles'
+    result = await _discord_request(url, tok)
+
+    json_s = result.json()
```
EarthenSky, line 150: I notice you sometimes use `json_s` and sometimes `result_json`, let's pick one to be consistent

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+
+async def get_all_roles(
+    id: str = guild_id
+) ->  dict[str, list[str]]:
+    tok = os.environ.get('TOKEN')
+    url = f'https://discord.com/api/v10/guilds/{id}/roles'
+    result = await _discord_request(url, tok)
+
+    json_s = result.json()
+    roles = [([role['id'], [role['name'], role['permissions']]]) for role in json_s]
```
EarthenSky, line 172: Should we make a Role dataclass l.ike with the others?

src/disc/disc.py
DerpyWasHere: Change made.

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+    return dict(roles)
+
+async def get_guild_members_with_role(
```
EarthenSky, line 175: I notice you sometimes use the terminology `user` and sometimes `guild_member`. Let's be consistent with terminology, but either is fine with me!

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+    url = f'https://discord.com/api/v10/guilds/{id}/members/{uid}'
+    result = await _discord_request(url, tok)
+
+    json_s = result.json()
+    return json_s['roles']
+
+async def get_all_roles(
+    id: str = guild_id
+) ->  dict[str, list[str]]:
+    tok = os.environ.get('TOKEN')
```
EarthenSky, line 167: add docstring explaining this function gets all roles in the specified guild

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+async def search_user(
+    user: str,
+    id: str = guild_id
+) -> User:
+    tok = os.environ.get('TOKEN')
+    url = f'https://discord.com/api/v10/guilds/{id}/members/search?query={user}'
+    result = await _discord_request(url, tok)
+    json = result.json()[0]['user']
+    user = User(json['id'], json['username'], json['discriminator'], json['global_name'], json['avatar'])
+    return user
```
EarthenSky, line 277: We should avoid shadowing the user parameter, as it could be confusing.
```suggestion
    return User(json['id'], json['username'], json['discriminator'], json['global_name'], json['avatar'])
```

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+    return channels
+
+async def search_user(
+    user: str,
+    id: str = guild_id
+) -> User:
+    tok = os.environ.get('TOKEN')
+    url = f'https://discord.com/api/v10/guilds/{id}/members/search?query={user}'
+    result = await _discord_request(url, tok)
+    json = result.json()[0]['user']
```
EarthenSky, line 275: will this function fail with IndexException (or something) if there are no user matches? We should return None if there is no match.

EarthenSky (changes requested):
Some small changes.

Also, I wasn't able to find the functions that does "get all users with access to a specific channel". Maybe you should add a docstring to those functions explaining that's what they do?

src/discord/discord.py
```diff
@@ -0,0 +1,277 @@
...
+
+DISCORD_CATEGORY_ID = 4
+ADMINISTRATOR = 0b1000
+VIEW_CHANNEL = 0b0010_0000_0000
+
+@dataclass
+class User:
+    id: str
+    username: str
+    discriminator: str
```
EarthenSky, line 18:
```suggestion
    # Discriminators are what used to be the #xxxx after a discord username. Accounts which haven't
    # migrated over yet have them still.
    discriminator: str
```

EarthenSky (changes requested):

src/discord/discord.py
EarthenSky: I have realized that users and guild_members are different!

src/discord/discord.py
DerpyWasHere: Handled in 9055fdf1e1c22c79984ea6a9278f3848b6ae0e53

src/discord/discord.py
DerpyWasHere: Handled in adeb09b55a4380b2458ddf0132ef924c1d20b612 and c8497bc28bc6b65b07b27dc5a05668839be42538.

src/discord/discord.py
DerpyWasHere: Handled in dbff44eeaf1f0663eea128a7a1c488cfc43ff158.

src/discord/discord.py
DerpyWasHere: Eeeeh its a bit simple, but we could.

src/discord/discord.py
DerpyWasHere: You're right. Fixed in 902d30c7bef9ff66caa2567984ae858f91d6d31f.

src/discord/discord.py
DerpyWasHere: Resolved in d34c09bf4e9d96944f9aa8920c99020b110b7e36.

EarthenSky (approved):
Woooo, hopefully it works! We'll have to setup testing later.

src/discord/discord.py
DerpyWasHere: Lets push this to another quick issue I think.

Issue #34, opened 2024-07-12
Rename officers/officers to officers/endpoints


EarthenSky:
The only files with routes should be called `endpoints.py` do this for all modules with routes. Optionally, do we want to call it `routes.py`, or even `urls.py` like django?

PR #45, opened 2024-07-25
Rename routes to url & fix bugs

EarthenSky:
Close https://github.com/CSSS/csss-site-backend/issues/34


EarthenSky:
"move fast and break things"

DerpyWasHere (commented):
No issues I can see. Could I get a quick rundown on why we renamed the routes though?

EarthenSky:
doing `import auth from auth` was a bit confusing, since it's implied that routes were at the root element of the module (pun not intended). making a file names urls.py makes it more clear where to find routing & that only routes should be in that file. This also helps avoid circular dependencies if a file gets too big.

Issue #42, opened 2024-07-15
Add limit to number of discord API requests per second


EarthenSky:
The discord API has a limit per second of 50 API requests. Unfortunately, many actions we want to take requires more than a single API request, and so a single user spamming a button might result in us hitting this limit.

Of course, the frontend should be careful with calling endpoints too quickly, but we should put a safeguard to stop this. A python threading lock + a counter to see how many API calls have been made every second would be a nice addition. If a 51st request is about to be made, sleep for 1s to avoid the penalty (or however long to not hit the discord limit).

There is also a limit per 10 minutes, which we don't want to hit either.

DerpyWasHere:
On any request, Discord's API returns headers which contain the following fields:

- X-RateLimit-Limit (Total number of requests that can be made)
- X-RateLimit-Remaining (Number of requests left before returning 429)
- X-RateLimit-Reset (Time at which the remaining number of requests resets in epoch time)
- X-RateLimit-Reset-After (Time remaining until the number of requests resets)
- X-RateLimit-Bucket (String identifying what resource is being rate limited)

These would be ideal to base our limiting behaviour on.

EarthenSky:
oh beautiful, so we shouldn't even need to do anything too fancy!

PR #47, opened 2024-08-03
Add rate limiting to Discord api requests

DerpyWasHere:
Added rate limiting to `_discord_request()` as described in issue #42.

A few design choices we could talk about:
- The constant we compare against
  - I chose 2 in case we have multiple threads running for some reason, we could choose differently. I found that the bucket size was 10 requests per 10 seconds.
- Sleep time
  - I chose to sleep for the remaining time until reset, but this might introduce blocking behavior we may not want. We could potentially handle this in some sort of priority queue? I think in our case though sleeping for the remaining time is sufficient as we're likely not splitting hairs over ms level behavior.
  - Alternatively this doesn't even matter as all of the functions in the Discord module are asynchronous anyway and expect to be waiting.


EarthenSky:
> The constant we compare against

What does 2 mean? 2 seconds to sleep after getting rate limited?

EarthenSky:
> I chose to sleep for the remaining time until reset, but this might introduce blocking behavior we may not want. We could potentially handle this in some sort of priority queue? I think in our case though sleeping for the remaining time is sufficient as we're likely not splitting hairs over ms level behavior.

Do you think this would be worthwhile to implement in the future? If so, let's make an issue for it. I'm happy with extra delay for the time being.

src/discord/discord.py
```diff
@@ -46,6 +46,13 @@ async def _discord_request(
             'User-Agent' : 'DiscordBot (https://github.com/CSSS/csss-site-backend, 1.0)'
         }
     )
+    rate_limit_reset = float(result.headers["x-ratelimit-reset-after"])
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+
+    if rate_limit_remaining <= 2:
+        from time import sleep
```
EarthenSky, line 53: please import this function at the top of the module. This avoids the function being reimported every time the rate limit is reached. It will only be imported once ever if at top of module!

DerpyWasHere:
> The constant we compare against
> > What does 2 mean? 2 seconds to sleep after getting rate limited?
2 seconds.

DerpyWasHere:
> > I chose to sleep for the remaining time until reset, but this might introduce blocking behavior we may not want. We could potentially handle this in some sort of priority queue? I think in our case though sleeping for the remaining time is sufficient as we're likely not splitting hairs over ms level behavior.

> Do you think this would be worthwhile to implement in the future? If so, let's make an issue for it. I'm happy with extra delay for the time being.

Honestly, not really. I doubt we'll hit a large enough scale to warrant it.

src/discord/discord.py
```diff
@@ -46,6 +46,13 @@ async def _discord_request(
             'User-Agent' : 'DiscordBot (https://github.com/CSSS/csss-site-backend, 1.0)'
         }
     )
+    rate_limit_reset = float(result.headers["x-ratelimit-reset-after"])
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
```
EarthenSky, line 50: is this number of remaining requests? if so, can you rename to `rate_limit_remaining_requests`, i see now why the number is 2. If there were 10 threads and we accidentally got rate limited, would this entry be 0, or would it not exist?

DerpyWasHere: The headers change completely iirc

EarthenSky (commented):
Looks good, just a few questions & requests!

Love how this is only a few lines of code. Mad respect.

src/discord/discord.py
DerpyWasHere: Resolved in ea025d3

src/discord/discord.py
DerpyWasHere: Resolved in f777c12

EarthenSky: > The headers change completely iirc

if the headers change completely, then this function would raise an exception for failing to index the x-ratelimit-remaining entry? Maybe we should handle that exception with a custom error?

DerpyWasHere: Ah shoot, I was looking at the wrong PR.  The headers do not change.

EarthenSky (approved):

PR #48, opened 2024-08-03
Github API interface with csss-site-backend

### LINKED ISSUE #20: control access of users to github
execs need to be automatically be added and/or removed from the github organization

You'll have to do some research into github's http api. Doing it async like with discord should be fine for now! If you can think of a better way to do it that would take a while, please make a new issue!

requirements:
- functions to add and remove github users from an org
- functions to add users to & from teams in a github org
- a module, just like the discord one
- vibe with us about what constraints the api has

DerpyWasHere:
Closes #20. (Vibes)

The REST API provided by Github has constraints detailed [here](https://docs.github.com/en/rest/orgs/members?apiVersion=2022-11-28). Realistically the only constraint _we_ care about is `x-ratelimit-remaining` which I've already added a check for. We should probably further discuss _how_ to gracefully handle this as currently I raise an exception when we hit <50 remaining API calls. As noted in #20, all of these functions are handled asynchronously, so I chose to leave a healthy margint o avoid being rate limited.

I added the functionality requested, adding/removing from organizations, adding/removing from teams. This is implemented in a new `github` module.

For testing I've created a new CSSS-Test-Organization, we can talk about transferring ownership of that in person/over call.

Authorization here in the `github` module is token based and administrated through the fine grained personal access token system in Github. As such we can just store it as an environment variable `GITHUB_TOKEN`.

### DIFF
--- a/src/github/github.py
+++ b/src/github/github.py
@@ -24,6 +24,10 @@ async def _github_request_get(
             "X-GitHub-Api-Version": "2022-11-28"
         }
     )
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+    if(rate_limit_remaining) < 50:
+        raise Exception("Less than 50 api calls remaining before being rate limited, please try again later")
+    
     return result
 
 async def _github_request_post(
@@ -40,6 +44,27 @@ async def _github_request_post(
         },
         data=post_data
     )
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+    if(rate_limit_remaining) < 50:
+        raise Exception("Less than 50 api calls remaining before being rate limited, please try again later")
+
+    return result
+
+async def _github_request_delete(
+        url: str,
+        token: str
+) -> Response:
+    result = requests.delete(
+        url,
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {token}",
+            "X-GitHub-Api-Version": "2022-11-28"}
+    )
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+    if(rate_limit_remaining) < 50:
+        raise Exception("Less than 50 api calls remaining before being rate limited, please try again later")
+
     return result
 
 async def get_user_by_username(
--- a/src/github/github.py
+++ b/src/github/github.py
@@ -124,19 +124,21 @@ async def get_user_by_id(
 
 async def add_user_to_org(
         org: str = github_org_name,
+        # We need one of either uid or email. We'll fail if we're provided both.
         uid: str | None = None,
         email: str | None = None
 ) -> None:
+    result = None
     if uid is None and email is None:
         raise ValueError("uid and username cannot both be empty")
-    result = None
+    elif uid is not None and email is not None:
+        raise ValueError("cannot populate both uid and email")
     # Arbitrarily prefer uid
-    if uid is not None and email is None:
+    elif uid is not None:
         result = await _github_request_post(f"https://api.github.com/orgs/{org}/invitations",
                                os.environ.get("GITHUB_TOKEN"),
                                dumps({"invitee_id":uid, "role":"direct_member"}))
-    # Assume at this point both exist, but use email
-    else:
+    elif email is not None:
         result = await _github_request_post(f"https://api.github.com/orgs/{org}/invitations",
                                os.environ.get("GITHUB_TOKEN"),
                                dumps({"email":email, "role":"direct_member"}))
--- a/src/github/github.py
+++ b/src/github/github.py
@@ -70,6 +70,11 @@ async def _github_request_delete(
 async def get_user_by_username(
         username: str
 ) -> GithubUser:
+    """
+        Takes in a Github username and returns an instance of GithubUser.
+
+        May return an empty list if no such user was found.
+    """
     result = await _github_request_get(f"https://api.github.com/users/{username}",
                               os.environ.get("GITHUB_TOKEN"))
     result_json = result.json()
@@ -78,6 +83,11 @@ async def get_user_by_username(
 async def get_user_by_id(
     id: str    
 ) -> int:
+    """
+        Takes in a Github user id and returns an instance of GithubUser.
+
+        May return an empty list if no such user was found.
+    """
     result = await _github_request_get(f"https://api.github.com/user/{id}",
                               os.environ.get("GITHUB_TOKEN"))
     result_json = result.json()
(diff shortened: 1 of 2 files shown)


DerpyWasHere:
Note that when generating organization invites, iirc we're restricted to 50/hr, so in _some_ scenarios we may want to implement rate limiting on the secondary rate limit (which is described on the rate limit page above)

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+    if result.status_code != 200:
+        raise Exception(f"Status code {result.status_code} returned when attempting to add user to team: {result_json['message']}")
+
+async def remove_user_from_team(
+        username: str,
+        slug: str,
+        org: str = github_org_name
+) -> None:
+    result = await _github_request_delete(f"https://api.github.com/orgs/{org}/teams/{slug}/memberships/{username}",
+                                          os.environ.get("GITHUB_TOKEN"))
```
EarthenSky, line 186: Not sure how you feel about it, but I think this style of line-break formatting is easier to read
```suggestion
    result = await _github_request_delete(
        f"https://api.github.com/orgs/{org}/teams/{slug}/memberships/{username}",
        os.environ.get("GITHUB_TOKEN"),
    )
```
(I found this in the python google style guide as well)

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+) -> GithubUser:
+    """
+        Takes in a Github username and returns an instance of GithubUser.
+
+        May return an empty list if no such user was found.
+    """
+    result = await _github_request_get(f"https://api.github.com/users/{username}",
+                              os.environ.get("GITHUB_TOKEN"))
+    result_json = result.json()
+    return [GithubUser(user["login"], user["id"], user["name"]) for user in [result_json]]
```
EarthenSky, line 110: What is `[result_json]` doing? I might suggest list(result_json) instead, unless you're making the list that contains only result_json, in which case maybe bind `user = result_json` instead.

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+
+async def add_user_to_org(
+        org: str = github_org_name,
+        uid: str | None = None,
+        email: str | None = None
+) -> None:
+    if uid is None and email is None:
+        raise Exception("uid and username cannot both be empty")
```
EarthenSky, line 131: Let's raise `ValueError` instead (since the value of uid and email is wrong)

src/github/github.py
DerpyWasHere: I see what you mean now though.

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+    result = None
+    # Arbitrarily prefer uid
+    if uid is not None and email is None:
```
EarthenSky, line 134: This is my pet peeve, feel free to disregard if you wish
```suggestion
    elif uid is not None and email is None:
```

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    result = await _github_request_get(f"https://api.github.com/user/{uid}",
+                              os.environ.get("GITHUB_TOKEN"))
+    result_json = result.json()
+    return [GithubUser(user["login"], user["id"], user["name"]) for user in [result_json]]
+
+async def add_user_to_org(
+        org: str = github_org_name,
+        uid: str | None = None,
+        email: str | None = None
```
EarthenSky, line 128: I would appreciate a quick docstring saying that at least one of these is required to call this function

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+) -> None:
+    if uid is None and email is None:
+        raise Exception("uid and username cannot both be empty")
+    result = None
+    # Arbitrarily prefer uid
+    if uid is not None and email is None:
+        result = await _github_request_post(f"https://api.github.com/orgs/{org}/invitations",
+                               os.environ.get("GITHUB_TOKEN"),
+                               dumps({"invitee_id":uid, "role":"direct_member"}))
+    # Assume at this point both exist, but use email
```
EarthenSky, line 138: let's fail if both are provided & put that in the docstring. I can forsee a subtle bug where someone accidentally provides the wrong email, but is confused because the uid is correct.

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+        url,
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {token}",
+            "X-GitHub-Api-Version": "2022-11-28"
+        }
+    )
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+    if rate_limit_remaining < 50:
+        raise Exception("Less than 50 api calls remaining before being rate limited, please try again later")
```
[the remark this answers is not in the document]
DerpyWasHere: Good call, 0d5f24b

src/github/github.py
```diff
@@ -0,0 +1,188 @@
...
+        username: str,
+        org: str = github_org_name
+) -> None:
+    if username is None:
+        raise Exception("Username cannot be empty")
+    result = await _github_request_delete(f"https://api.github.com/orgs/{org}/memberships/{username}",
+                                          os.environ.get("GITHUB_TOKEN"))
+    # Logging here potentially?
+    if result.status_code != 204:
+        raise Exception(f"Status code {result.status_code} returned when attempting to delete user {username} from organization {org}")
```
EarthenSky, line 158: I'm hoping to log all exceptions that bubble up from an endpoint when a function is called. However, if we plan to handle these kinds of errors, then I would prefer to not use exceptions & log at a higher level. It's probably fine for now, but we might want to refactor a lil depending on how implementation goes.

src/github/github.py
DerpyWasHere: Fixed in ed749de

EarthenSky (commented):
Love the comments & how similar this is to the discord module. Damn, you're going so fast, this is lovely!

How are you testing these functions?

We should have some docs on how to do testing for these modules? Not sure, maybe that's for later.

src/github/github.py
DerpyWasHere: Resolved in be0659a

src/github/github.py
DerpyWasHere: Resolved in be0659a

DerpyWasHere:
I'm testing this on a test organization `CSSS-Test-Organization` that I made. Albeit far from professional I've just been using a test python script.

src/github/github.py
DerpyWasHere: Yea, I want to replace these exceptions with logging calls, they're just here so they're blatantly obvious that we can replace them with logging calls.

src/github/github.py
```diff
@@ -0,0 +1,192 @@
...
+        headers={
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {token}",
+            "X-GitHub-Api-Version": "2022-11-28"
+        },
+        data=post_data
+    )
+    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
+    if rate_limit_remaining < 50:
+        return None
```
EarthenSky, line 59: Remember to update the return type if it's optional
```suggestion
) -> Response | None:
    result = requests.post(
        url,
        headers={
            "Accept": "application/vnd.github+json",
            "Authorization": f"Bearer {token}",
            "X-GitHub-Api-Version": "2022-11-28"
        },
        data=post_data
    )
    rate_limit_remaining = int(result.headers["x-ratelimit-remaining"])
    if rate_limit_remaining < 50:
        return None
```

src/github/github.py
EarthenSky: this is good, but fyi docstrings ```"""this is a docstring"""``` as the first line of a function is picked up by the intellisense.

src/github/github.py
EarthenSky: yee!

src/github/github.py
DerpyWasHere: 5327538

src/github/github.py
DerpyWasHere: Resolved in 0ddf93e

src/github/github.py
DerpyWasHere: Ah you're right. 29c04e2

EarthenSky (approved):

PR #51, opened 2024-08-05
Exam Bank Watermarking

### LINKED ISSUE #12: insert watermarks over pdfs in real time
The watermark should be tiled diagonally with the current user's computing id + "This exam can only be viewed by SFU Professors"

The watermark should be rasterized over the image.

The blitting of the watermark shouldn't take more than 1 second for a pdf with 10 pages, else it will feel sluggish.

This should be implemented as a python function and a test page.

DerpyWasHere:
Closes #12.

Created an exambank module which exposes 3 functions:

- create_watermark() -> BytesIO
- apply_watermark() -> BytesIO
- raster_pdf() -> BytesIO

These three functions take a given exam, watermarks it, and rasterizes it to prevent the removal of said watermark.

Improvements could be to remove the usage of the pypdf library as pymupdf may be able to perform the same functionality faster. Benchmarking could be useful here. This could also be expressed as a new issue for later down the road.


EarthenSky:
I changed the target branch to make reviewing eaiser, make sure to change it back to main before merging

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+from pymupdf import Matrix
+from pymupdf import open as p_open
+from pypdf import PdfReader, PdfWriter
+from reportlab.lib import colors
+from reportlab.lib.pagesizes import A4
+from reportlab.lib.units import inch
+from reportlab.pdfgen import canvas
+
+# param to describe margin for exam generation text
+border = 20
```
EarthenSky, line 13: let's use SCREAM CASE FOR CONSTANTS
```suggestion
BORDER = 20
```

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+
+    warning_pdf.save()
+    warning_buffer.seek(0)
+
+    # Merge into watermark
+    watermark_pdf = PdfWriter()
+    stamp_pdf = PdfReader(stamp_buffer)
+    warning_pdf = PdfReader(warning_buffer)
+    # Destructively merges in place
+    stamp_pdf.pages[0].merge_page(warning_pdf.pages[0])
```
EarthenSky, line 52: What is a page merge? Does this overwrite the old page?

DerpyWasHere: Yes. This overwrites the copy in memory. In this case, it'd be the copy read into stamp_pdf.

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+from reportlab.lib.units import inch
+from reportlab.pdfgen import canvas
+
+# param to describe margin for exam generation text
+border = 20
+
+def create_watermark(
+        computing_id: str,
+        density: int = 5
+) -> BytesIO:
```
EarthenSky, line 18: can you add a docstring saying that the bytes object returned is a PDF with a single page, containing the watermark image? (if I understood correctly)

Additionally, would it make sense to return an image only, and not a pdf containing an image?

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+    watermark_pdf.add_page(stamp_pdf.pages[0])
+
+    watermark_buffer = BytesIO()
+    watermark_pdf.write(watermark_buffer)
+    watermark_buffer.seek(0)
+    return watermark_buffer
+
+def apply_watermark(
+        pdf_path: Path | str,
+        # expect a BytesIO instance (at position 0), accept a file/path
```
EarthenSky, line 65: let's not try to accept several different kinds of inputs, but only choose one. This makes the implementation less complicated. Let's just take bytes io. If we want to take a path, let's make a function  `watermark_from_path` or something similar.

src/exambank/exambank.py
DerpyWasHere: Sure, resolved in 734f92a.

We could do that I think, that would require us to turn it into an image first though. Doing this might make it easier to replace pypdf with pymupdf later on.

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+        stamp: BytesIO | Path | str,
+) -> BytesIO:
+    # process file
+    stamp_page = PdfReader(stamp).pages[0]
+
+    writer = PdfWriter()
+    reader = PdfReader(pdf_path)
+    writer.append(reader)
+    for pdf_page in writer.pages:
+        pdf_page.transfer_rotation_to_content()
```
EarthenSky, line 75: What does this function do?

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
+from io import BytesIO
+from pathlib import Path
+
+from pymupdf import Matrix
+from pymupdf import open as p_open
```
EarthenSky, line 5: I would prefer doing import pymupdf, then `pymupdf.open(...)`, but it's up to you~!

src/exambank/exambank.py
DerpyWasHere: Sure, resolved in 6ff28fd

src/exambank/exambank.py
DerpyWasHere: If you don't do the transfer_rotation_to_content() call, the watermark is applied upside down.

src/exambank/exambank.py
```diff
@@ -0,0 +1,112 @@
...
+def raster_pdf(
+        pdf_path: BytesIO | Path | str,
+        dpi: int = 300
+) -> BytesIO:
+    raster_buffer = BytesIO()
+    # adapted from https://github.com/pymupdf/PyMuPDF/discussions/1183
+    if isinstance(pdf_path, BytesIO):
+        with p_open(stream=pdf_path) as doc:
+            page_count = doc.page_count
+            with p_open() as target:
```
EarthenSky: what does p_open do with no parameters? I assume it would try to open a file, but no file is provided...

EarthenSky (changes requested):
How did you test this?

I don't want to force you to make unit or integration tests, but it would be nice to include the scripts you use to test it, just so that I can see it worked on your machine.

src/exambank/exambank.py
DerpyWasHere: I'm uh, not exactly sure.

src/exambank/exambank.py
DerpyWasHere: Sure, 5f9895b.

DerpyWasHere:
I've been using a really scuffed script that I intermittently comment out things in :sob:

EarthenSky:
They don't need to be good tests, just "works on my machine" proof

src/exambank/exambank.py
EarthenSky: \<3

src/exambank/exambank.py
EarthenSky: OH! interesting. I guess since PDF is composed of elements, there are some transformations that are associated with the elements? Odd that upside down is the default....

src/exambank/exambank.py
EarthenSky: hmmm, well let's at least have a bit of an understanding of what it does

src/exambank/exambank.py
DerpyWasHere: Yeahhh. It's mentioned in their documentation that that may happen.

src/exambank/exambank.py
DerpyWasHere: Okay, I sniffed through the source code a bit. open is an alias for the Document class, so we're just constructing an instance of Document which in turn is a blank PDF.

EarthenSky: woo! Makes a lot of sense now

EarthenSky (changes requested):
Oop, small think I just noticed, can we rename `exambank/exambank.py` to `exambank/__init__.py`? This puts the functions in the root of the exambank module, so we can do `from exambank import create_watermark`

DerpyWasHere:
Reflected in 31df22b!

DerpyWasHere (commented):
Handled all requests.

EarthenSky (approved):
merged changes from github PR

PR #61, opened 2024-08-07
Fix alembic

EarthenSky:
I haven't read enough alembic documentation and made a mistake, more reading I shall do!

In the meantime, this should fix things enough for others to work with it.

This commit merges the two heads from separate PRs (mine & micahs), so that new developers can apply all migrations just by doing `alembic upgrade head`. I'm not sure the correct way to usually go about this, but I guess I should have rebased over his changes & done a merge? Or even maybe an alembic rebase since our changes were unrelated.


DerpyWasHere (commented):
Looks fine to me, we should have a quick chat at some point catching me up on the changes though.

Issue #54, opened 2024-08-05
CSSS Blog backend


EarthenSky:
We wanted a blog, so we need to add support for it.

I imagine we'll want to store everything in a database table, then dynamically serve content based on the page.

This task will have several steps:
- [ ] Create a table in the database to store blog entries 
    - this table should store:
- [ ]
- [ ] Migrate all past

EarthenSky:
As always, if you need help with anything or with clarifications, feel free to ask me anything!!

gunmack:
1. For storing computing IDs:
              Should the data type be:
              - Text
              - Varchar(n)
              
2. For storing html content as a string:
              - Assuming the content of a post would generally be large, is it okay to use [TOAST](https://www.postgresql.org/docs/current/storage-toast.html) compression method?
              
3. For storing tags:
             - In case the list of tags is large, can TOAST be used?

EarthenSky:
> 1. For storing computing IDs

You can find the size used in the existing database tables [here](https://github.com/CSSS/csss-site-backend/blob/main/src/auth/models.py) => we use Varchar([16](https://github.com/CSSS/csss-site-backend/blob/44a7088479b814dc96f1fc9e1e3128454704c00c/src/constants.py#L8-L9)). It's a little overkill just in case the size increases in the future, but SFU claims there is a max length of 9 chars (this might include non-ascii wide chars, so buffer space is good to have).

> 2. For storing html content as a string

For the first version, we want a quick & easy proof of concept. No need to make something optimal before profiling / evaluating it. Once the simple version works good, we can upgrade to using something like TOAST. I've never heard of it before, but if you think it's easy to implement, go for it!

That being said, tell me about how TOAST works, because it looks like psql does it automatically? => "large field values are compressed and/or broken up into multiple physical rows ... This happens transparently to the user ... The technique is affectionately known as TOAST"

> 3. For storing tags

Before we consider things like TOAST, please take a look at how authentication is implemented in our stack using alembic & sqlalchemy. I know you're interested in psql, but we might be working at a higher level for the time being so we might not get to do super cool things ; _ ;

That being said, i doubt we'll have more than 8kb of text for tags, so no TOAST if it's not automatic. Though, my assumption was that having an entry with a large amount of text would only reduce retrieval performance. Do you think this 8kb limit might truncate content stored?

// Aside:

[...]

PR #62, opened 2024-08-09
Blog table model

gunmack:
For https://github.com/CSSS/csss-site-backend/issues/54

- Added a model for the blog_posts table


src/alembic/env.py
```diff
@@ -42,6 +43,7 @@ def run_migrations_offline() -> None:
     # url = config.get_main_option("sqlalchemy.url")
     context.configure(
         url=database.SQLALCHEMY_DATABASE_URL,
+        # url=database.SQLALCHEMY_TEST_DATABASE_URL,
```
EarthenSky, line 46: ah! This is a good point; we'll want to add support for testing migrations using the local test database.

Though actually, the only reason for the test database is for running things like unit & integration tests, which can freely assume the most up to data DB, so for the time being feel free to just use the main database for tests!

src/blog/models.py
```diff
@@ -0,0 +1,30 @@
...
+    __tablename__ = "blog_posts"
+
+    # title of post ( meant to be an unique key but for simplicity,
+    # using already defined primay_key )
+    title = Column(String(128), primary_key=True, nullable=False)
+
+    # computing id
+    # TODO: add foreign key
+    computing_id = Column(String(COMPUTING_ID_LEN),
+        # ForeignKey(),
```
EarthenSky, line 18: Oh, just as an fyi, sqlalchemy has support for something called a `relationship` to simplify the access pattern of tables related by foreign keys. Due to restrictions of the library with async databases, we're planning on not using this feature.

Mentioning it as you might see the feature in the documentation.

micahdbak: yea relationships be silly

Implementing foreign keys using equality is sorta bad practice but who cares, it's simple, maybe just do that for now 😇 

I.e., select from another table where computing_id = this.

EarthenSky: Do you mean bad practice with sqlalchemy? Do you know why? 

It actually kinda simplifies creating alembic migrations since alembic doesn't have support for relationships, which might confuse some people (since relationships aren't even stored in the schema).

micahdbak: it's bad practice in general, for databases, to not use foreign keys when the actual use of a column is that of a foreign key. I.e. you could make an insertion that is invalid and it wouldn't throw an error

BUT if you're aware of these impacts and intend to mitigate them programmatically, who cares :))

micahdbak: I'm referring to foreign keys / database constructs in general rather than alembic relationship BTW - that seems to be an oddity that i dont exactly like LMAO

DerpyWasHere: Yea I was expecting to see a ForeignKeyConstraint here

DerpyWasHere (commented):
Looks fine to me so far, potentially take a look into sqlalchemy ForeignKeyConstraints.

src/alembic/versions/ce9a59922b43_tags_post_tags.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision: str | None = "e398258d0c63"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("blog_posts", sa.Column("post_tags", sa.String(length=128), nullable=True))
+    op.drop_column("blog_posts", "tags")
+    # ### end Alembic commands ###
```
EarthenSky, line 25: let's combine these three migrations into just one for the tables (one migration per PR is the typical rule). Afterwards, double check that everything is the same as your `models.py`. Once that's the case, remove the commands "auto generated by Alembic"

Alembic will not do constraints correctly, so you'll have to make sure the ForeignKeyConstraint ids match manually. I'll double check afterwards though, so don't worry too much about it being perfect!

src/blog/crud.py
```diff
@@ -0,0 +1,65 @@
...
+
+async def create_new_entry(db_session: database.DBSession,
+                            title:str, computing_id:str, post_tags: str, html_content:str,
+                            date_created:date, last_edited: date):
+
+    """ To create a new blog entry """
+
+    return None
+
+async def fetch_by_title(db_session: database.DBSession, title:str) -> (str, str, datetime, list[str] | None) | None:
```
EarthenSky, line 23:
```suggestion
async def fetch_by_title(db_session: database.DBSession, title:str) -> tuple[str, str, datetime, list[str] | None] | None:
```
unfortunately tuple type hints are not formatted the obvious way (I hope they switch it one day, but I doubt it)

src/blog/url.py
```diff
@@ -0,0 +1,15 @@
...
+import database
+from fastapi import APIRouter, Request
+from fastapi.responses import JSONResponse
+from permission.types import OfficerPrivateInfo
+
+_logger = logging.getLogger(__name__)
+
+router = APIRouter(
+    prefix="/blog_entries",
+    tags=["blog_entries"],
```
EarthenSky, line 14: let's make the prefix /blog (& tag blog), as mentioned in the issue

src/blog/models.py
EarthenSky, line 1: let's rename this file to `tables.py`. Not your fault, I only just decided to change the naming convention a few days ago (because "models" is not very clear, when pydantic models exist... although we aren't using them, tables is clearer)

EarthenSky (changes requested):
Looks good! Here are some small suggestions. I'll double check if the ForeignKey is formatted right later tonight (I need to create foreign keys for the old tables).

I think once the blog table seems good, we don't need to implement the crud or urls.py any further.

You'll just want to clean up the formatting & comments, then we're good!

src/alembic/env.py
EarthenSky: WAIT, no we won't want to add support for migrations to the test database!

I totally forgot why I created a test database in the first place, so I've added it to our FAQ here https://github.com/CSSS/csss-site-backend/wiki/FAQ#6-why-is-there-a-main-database--a-test-database

The reason is that the test database is for use without migrations, so we can do testing more quickly (it ignores alembic & just wipes & recreates the database whenever `./load_test_db.py` is run)

The main database is for use with migrations.

src/alembic/env.py
```diff
@@ -58,6 +59,7 @@ async def run_async_migrations() -> None:
     """
     configuration = config.get_section(config.config_ini_section)
     configuration["sqlalchemy.url"] = database.SQLALCHEMY_DATABASE_URL
+    # configuration["sqlalchemy.url"] = database.SQLALCHEMY_TEST_DATABASE_URL
```
EarthenSky, line 62: the purpose of the test db is to run tests from scratch (without migrations), so we should remove this comment as we don't want to accidentally do so

src/blog/tables.py
```diff
@@ -0,0 +1,31 @@
...
+class BlogPosts(Base):
+    # table name
+    __tablename__ = "blog_posts"
+
+    # title of post ( meant to be an unique key but for simplicity,
+    # using already defined primay_key )
+    title = Column(String(128), primary_key=True, nullable=False)
+
+    # computing id
+    # TODO: add foreign key
```
EarthenSky, line 17:
```suggestion
```

we can remove this TODO now \:)

src/alembic/versions/2a6ea95342dc_blog_posts.py
```diff
@@ -0,0 +1,39 @@
...
+    op.create_table('blog_posts',
+    sa.Column('title', sa.String(length=128), nullable=False),
+    sa.Column('computing_id', sa.String(length=32), nullable=False),
+    sa.Column('date_created', sa.DateTime(), nullable=False),
+    sa.Column('last_edited', sa.DateTime(), nullable=False),
+    sa.Column('html_content', sa.Text(), nullable=False),
+    sa.Column('post_tags', sa.String(length=128), nullable=True),
+    sa.ForeignKeyConstraint(['computing_id'], ['officer_info.computing_id'], ),
+    sa.PrimaryKeyConstraint('title')
+    )
```
EarthenSky, line 32:
```suggestion
    op.create_table('blog_posts',
        sa.Column('title', sa.String(length=128), primary_key=True, nullable=False),
        sa.Column('computing_id', sa.String(length=32), sa.ForeignKey("officer_info.computing_id"), nullable=False),
        sa.Column('date_created', sa.DateTime(), nullable=False),
        sa.Column('last_edited', sa.DateTime(), nullable=False),
        sa.Column('html_content', sa.Text(), nullable=False),
        sa.Column('post_tags', sa.String(length=128), nullable=True),
    )
```

see: https://github.com/CSSS/csss-site-backend/pull/78/files#diff-98a11339635fcc0df585c8fc64cd05ca369d9994dd56dd0317447475904854a6 for examples

src/alembic/versions/2a6ea95342dc_blog_posts.py
```diff
@@ -0,0 +1,39 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
EarthenSky, line 37: everything looks good in this migration! Let's remove these "###" comments

EarthenSky (changes requested):
All good! Just a few small changes which I can do myself :)

Do you have anything else to add or is it good to merge?

EarthenSky (approved):

gunmack:
That should be all. 
Regarding testing the table:
- I checked manually in the psql terminal to make sure that the table is functional

image

EarthenSky:
> That should be all

lovely! I'll merge it now

PR #63, opened 2024-08-09
First Elections PR

### LINKED ISSUE #25: Elections Info
We'll need database tables to store past elections, enable admins to appoint an elections officer, allow elections officer to configure an election, and to allow logged in student to run for executive positions.

TODO: add more issues for the above tasks

EarthenSky:
elections PR, woo!


src/elections/models.py
```diff
@@ -0,0 +1,62 @@
...
+    speech = Column(Text)
+
+    __table_args__ = (
+        PrimaryKeyConstraint(computing_id, nominee_election),
+        ForeignKeyConstraint(
+            ["computing_id"], ["election_nominee.computing_id"]
+        ),
+        ForeignKeyConstraint(
+            ["nominee_election"], ["election.slug"]
+        )
```
EarthenSky, line 61: There's a slightly nicer way to mark foreign keys, which we might want to use in the future, via `ForeignKey`. Though, I'm not entirely sure, so just FYI.

The reason is that we can create an automated naming convention, so we can't accidentally name constraints incorrectly. Take a quick look at https://alembic.sqlalchemy.org/en/latest/naming.html#:~:text=automated%20naming%20convention I'll add support for this soonish!

DerpyWasHere: Can we use ForeignKey for compound fks?

EarthenSky: I've decided automated naming convention will be confusing & error prone. Let's just name them ourselves \:) (or write a module if we really care)

src/elections/tables.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Previously named sfuid
+    computing_id = Column(String(COMPUTING_ID_LEN), primary_key=True)
+    full_name = Column(String(64), nullable=False)
+    facebook = Column(String(64))
+    instagram = Column(String(64))
+    email = Column(String(64))
+    discord = Column(String(DISCORD_NAME_LEN))
+    discord_id = Column(String(DISCORD_ID_LEN))
+    discord_username = Column(String(DISCORD_NICKNAME_LEN))
```
EarthenSky, line 46: I'm not sure how this info is being stored in the current website, but officers may be allowed to store new copies of this per-election. This begs the question; how backwards compatible do we want to be with the old site?

I think it would be nice to include all past information (html formatted speeches, etc...), but maybe for personal stuff like this we can just keep the most recent version. @micahdbak what do you think?

src/elections/tables.py
```diff
@@ -0,0 +1,58 @@
...
+# Each row represents an instance of an
+class Election(Base):
+    __tablename__ = "election"
+
+    # Slugs are unique identifiers
+    slug = Column(String(32), primary_key=True)
+    name = Column(String(32), nullable=False)
+    officer_id = Column(String(COMPUTING_ID_LEN), nullable=False)
+    # Can be one of (general_election: General Election, by_election: By-Election, council_rep_election: Council Rep Election)
+    type = Column(String(64), default="general_election")
```
EarthenSky, line 29: If you place the actual enum as the type, it will insert the enum's string into the db

see https://stackoverflow.com/questions/2676133/best-way-to-do-enum-in-sqlalchemy

Above comment no longer needed

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+
+router = APIRouter(
+    prefix="/elections",
+    tags=["elections"],
+)
+
+class ElectionType(Enum):
+    GENERAL_ELECTION = "general_election"
+    BY_ELECTION = "by_election"
+    COUNCIL_REP_ELECTION = "council_rep_election"
```
EarthenSky, line 30: I have a personal grudge against python enums which may or may not be reasonably motivated; they often only get converted into string, and code gets polluted with `x.value`. I bring this up especially since if their purpose is to be stored in the database, that involves converting them into string (in the backend).

It would be better to just have them be string constants in the first place. This approach has a few small negatives, but ultimately reduces complexity, as you never need to worry about whether an instance contains a string of the enum, or the enum itself.

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+)
+async def update_election(
+    request: Request,
+    db_session: database.DBSession,
+    slug: str,
+    name: str,
+    election_type: str,
+    date: datetime | None = None,
+    end_date: datetime | None = None,
+    websurvey: str | None = None
```
EarthenSky, line 132: Might be best to turn these into a single parameter using a dataclass. Not sure if it makes the `/api/docs` page look nicer, but it would be a bit cleaner.

EarthenSky (commented):
I'll do another review round after, but 99% looks good!

src/elections/crud.py
```diff
@@ -0,0 +1,68 @@
...
+from elections.tables import Election
+from officers.constants import OfficerPosition
+from officers.tables import OfficerInfo, OfficerTerm
+
+_logger = logging.getLogger(__name__)
+
+async def get_election(db_session: database.DBSession, election_slug: str) -> Election | None:
+    query = sqlalchemy.select(Election)
+    query = query.where(Election.slug == election_slug)
+    result = (await db_session.execute(query)).scalar()
```
EarthenSky, line 16:
```suggestion
    query = (
        sqlalchemy
        .select(Election)
        .where(Election.slug == election_slug)
    )
    result = await db_session.scalar(query)
```
a more sql way of formatting queries + we can skip the `.execute()` call

src/elections/crud.py
```diff
@@ -0,0 +1,68 @@
...
+    db_session.commit()
```
EarthenSky, line 17: Here's an interesting question: how do we know if a crud function has committed current db session or not? The unfortunate answer is that you have to read the function every time. 

For this reason, it's best to never commit in a crud function, so that the functions can be used in different contexts. Instead, transactions should be managed on a per-endpoint basis.

src/elections/crud.py
```diff
@@ -0,0 +1,68 @@
...
+    return result
+
+async def create_election(params: dict[str, datetime], db_session: database.DBSession) -> None:
```
EarthenSky, line 20: We should probably make a dataclass out of this `dict`, to be a little more strongly typed?

src/elections/tables.py
```diff
@@ -0,0 +1,58 @@
...
+    date = Column(DateTime, nullable=False)
+    end_date = Column(DateTime)
+    websurvey = Column(String(300))
```
EarthenSky, line 32: rip websurvey. Is this supposed to store a websurvey link? Since websurvey is dead now, maybe we should just make it `survey_link`?

src/permission/types.py
```diff
@@ -31,6 +33,21 @@ async def has_permission(db_session: database.DBSession, computing_id: str) -> b
 
         return most_recent_exec_term.end_date > cutoff_date
 
+class ElectionOfficer:
+    @staticmethod
+    async def has_permission(db_session: database.DBSession, computing_id: str) -> bool:
+        """
+        An current elections officer has access to all elections, prior elections officers have no access.
+        """
+        officer_terms = await officers.crud.current_executive_team(db_session, True)
```
EarthenSky, line 42: `current_executive_team` may enforce certain permissions in order to get private data from other people (TODO: double check this gabe). I created a function in https://github.com/CSSS/csss-site-backend/pull/86/commits/746e92eaffec3c9573eb17c9069b0e71fe505e02 to get the list of positions a single officer has, which would be better for this, though depends on my issue getting completed...

src/elections/crud.py
```diff
@@ -0,0 +1,68 @@
...
+    Updates an election with the provided parameters.
+    Take care as this will replace values with None if not populated.
+    You _cannot_ change the name or slug, you should instead delete and create a new election.
+    Does not validate if an election _already_ exists
+    """
+
+    election = (await db_session.execute(sqlalchemy.select(Election).filter_by(slug=params["slug"]))).scalar_one()
+
+    if params["date"] is not None:
+        election.date = params["date"]
```
EarthenSky, line 55: I'm not sure if this syntax actually emits an sql statement, but maybe we could use an update function?

src/elections/tables.py
micahdbak: > I think it would be nice to include all past information (html formatted speeches, etc...), but maybe for personal stuff like this we can just keep the most recent version. @micahdbak what do you think?

agree, let's just store the most up to date contact info

src/alembic/versions/243190df5588_create_election_tables.py
```diff
@@ -0,0 +1,61 @@
...
+
+# revision identifiers, used by Alembic.
+revision: str = "243190df5588"
+down_revision: str | None = "166f3772fce7"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
```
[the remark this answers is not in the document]
DerpyWasHere: Migration matches, will the constraint names not get automatically generated from the convention mentioned above?

src/elections/tables.py
DerpyWasHere: Sorry, I don't think I follow here. What do you mean by most recent version?

src/elections/tables.py
DerpyWasHere: Removed comment.

src/elections/urls.py
DerpyWasHere: I changed the enum into a plain string array. It removed a list comprehension that we probably didn't need in the first place in create_election().

src/elections/crud.py
DerpyWasHere: This emits an SQL statement and will just fetch the first matching slug (since slugs are supposed to be unique). An update isn't necessary as this isn't actually modifying anything.

src/elections/crud.py
DerpyWasHere: Ahhh I see what you're saying. I'll take out the commit call and delegate it to any endpoint that needs to call it.

DerpyWasHere: Ahhh, I see what you mean. I'll delegate committing to endpoints as you've suggested. This should be resolved in [e2bb4db](https://github.com/CSSS/csss-site-backend/pull/63/commits/e2bb4db03880bdf77cc203fcf40688ecea53dd5e).

src/elections/crud.py
DerpyWasHere: Done, see [82ccc24](https://github.com/CSSS/csss-site-backend/pull/63/commits/82ccc2414f96bc616ed7546ae2f1bbcea4a70f21).

src/permission/types.py
DerpyWasHere: I think it'd be more prudent to handle this after merging this PR into the main branch so that I can access the functions you've made in [746e92e](https://github.com/CSSS/csss-site-backend/commit/746e92eaffec3c9573eb17c9069b0e71fe505e02).

src/elections/tables.py
DerpyWasHere: Change is reflected in [50302b1](https://github.com/CSSS/csss-site-backend/pull/63/commits/50302b1ee67a2ae6b17a2c46bc62bd241d622888) and [82ccc24](https://github.com/CSSS/csss-site-backend/pull/63/commits/82ccc2414f96bc616ed7546ae2f1bbcea4a70f21).

src/elections/crud.py
```diff
@@ -0,0 +1,68 @@
...
+               name=params["name"],
+               officer_id=params["officer_id"],
+               type=params["type"],
+               date=params["date"],
+               end_date=params["end_date"],
+               websurvey=params["websurvey"])
+    db_session.add(election)
+    await db_session.commit()
+
+async def delete_election(slug: str, db_session: database.DBSession) -> None:
```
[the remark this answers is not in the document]
DerpyWasHere: Fixed in [a36eef6](https://github.com/CSSS/csss-site-backend/pull/63/commits/a36eef63b81220c9b737c0bb99a5dfb9e752b31e).

src/elections/crud.py
DerpyWasHere: It doesn't modify any data. It just grabs the first matching election corresponding to a slug. I use an update function later in the function.

DerpyWasHere:
Alright, brief high level summary. This PR introduces the following:
- CRUD functions for creating, getting, updating, and deleting elections.
- SQLAlchemy tables for elections, nominees, and nominee applications (election paired with nominee)
- Endpoints for creating, deleting, and updating elections.

This still needs to introduce/do the following:
- Endpoint for getting/viewing an election's data (public and private).
- Test user validation. 

I'll work on these in my next PR.

As of this comment, all of the above comments should have been addressed.

src/elections/urls.py
DerpyWasHere: I've done so in [e2bb4db](https://github.com/CSSS/csss-site-backend/pull/63/commits/e2bb4db03880bdf77cc203fcf40688ecea53dd5e).

src/alembic/versions/243190df5588_create_election_tables.py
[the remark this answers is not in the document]
EarthenSky: I think the constraint names will get automatically matched; however you may have noticed that some things like `default="general_election"` have not been carried over.

Once you've moved the extra things, please remove the `###` comments

src/permission/types.py
EarthenSky: if you merge main into this branch, you can access those functions. You've already done this, so you can use `current_officer_positions`

src/alembic/versions/243190df5588_create_election_tables.py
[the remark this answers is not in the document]
DerpyWasHere: I added back the default that didn't get carried over in [853038d](https://github.com/CSSS/csss-site-backend/pull/63/commits/853038d29fd44ce131fde467bce0fdac2f855241). Lemme know if I missed anything else.

src/permission/types.py
DerpyWasHere: Perfect, I'll get on that in my next PR then.

src/elections/tables.py
EarthenSky: ah, by "most recent version" I think I meant only new data inputted by our new backend. For example, sfucsss.org/elections data is stored in a db table originally created by Jace, which we want to eventually migrate here.

src/elections/tables.py
```diff
@@ -0,0 +1,59 @@
...
+# Each row represents an instance of an
+class Election(Base):
+    __tablename__ = "election"
+
+    # Slugs are unique identifiers
+    slug = Column(String(32), primary_key=True)
+    name = Column(String(32), nullable=False)
+    officer_id = Column(String(COMPUTING_ID_LEN), nullable=False)
+    type = Column(String(64), default="general_election")
+    date = Column(DateTime, nullable=False)
```
EarthenSky, line 31: is `date` the start date for the election? if so, `start_date` would be more clear.

In addition, since hours, minutes, seconds are important (speeches are usually made public at the same time as voting is allowed, which is done on a per-hour basis usually), should we have `start_datetime` and `end_datetime` to be more explicit?

src/elections/tables.py
```diff
@@ -0,0 +1,59 @@
...
+    __tablename__ = "election_nominee"
+
+    # Previously named sfuid
+    computing_id = Column(String(COMPUTING_ID_LEN), primary_key=True)
+    full_name = Column(String(64), nullable=False)
+    facebook = Column(String(64))
+    instagram = Column(String(64))
+    email = Column(String(64))
+    discord = Column(String(DISCORD_NAME_LEN))
+    discord_id = Column(String(DISCORD_ID_LEN))
```
EarthenSky, line 46: technically `DISCORD_ID_LEN` is not 18, it reached past 19 a while ago and is getting close to 20 iirc. I updated it to `32` in my last PR merged into main, so you should update it in the migration as well.

src/alembic/versions/243190df5588_create_election_tables.py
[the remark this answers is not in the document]
EarthenSky: looks good to me; I did find one small issue with the discord id being of size 18 but I made a comment for it.

src/elections/crud.py
EarthenSky: Does setting `election.date = datetime.now()` emit an SQL UPDATE? or does it just update the local variable. iirc if you do this assignment then commit, it should update the assignment without the following UPDATE.

src/elections/tables.py
DerpyWasHere: I've updated date -> start_datetime and end_date -> end_datetime in [57f4b2a](https://github.com/CSSS/csss-site-backend/pull/63/commits/57f4b2a7064cddb79560da80990507a2f3adf2b2).

src/elections/tables.py
DerpyWasHere: Resolved in [2dee83a](https://github.com/CSSS/csss-site-backend/pull/63/commits/2dee83a24b8858017ed260933a0767f60141a36e).

src/elections/crud.py
DerpyWasHere: Didn't you suggest above to _not_ commit at the end of this and let API endpoints handle the commit? The elections/update_election endpoint has the commit for this function.

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    Creates a unique slug based on text passed in. Assumes non-unicode text.
+    """
+    return re.sub(r"[\W_]+", "-", text)
+
+async def _validate_user(
+        db_session: database.DBSession,
+        session_id: str
+) -> dict:
+    computing_id = await auth.crud.get_computing_id(db_session, session_id)
```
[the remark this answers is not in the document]
EarthenSky: Also, why does validate user return dict, but you compare it against a bool later?

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+    user_auth = await _validate_user(db_session, session_id)
+    if user_auth is False:
+        # let's workshop how we actually wanna handle this
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="You do not have permission to access this resource",
+            headers={"WWW-Authenticate": "Basic"},
+        )
+
+    # Default start time should be now unless specified otherwise
```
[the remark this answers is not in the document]
EarthenSky: OR, we could require the start time to be specified on creation

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+    if start_datetime is None:
+        start_datetime = datetime.now()
+
+    if election_type not in election_types:
+        raise RequestValidationError()
+
+    params = ElectionParameters(
+        _slugify(name),
+        name,
+        await auth.crud.get_computing_id(db_session, session_id),
```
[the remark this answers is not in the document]
EarthenSky: note about naming: I had to have a distinction between the computing_id of the current session vs the computing_id of the person being looked up (you must be admin to lookup info about others, but you can about yourself)

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+
+    if slug is not None:
+        await elections.crud.delete_election(slug, db_session)
+        await db_session.commit()
+
+@router.get(
+    "/update_election",
+    description="""Updates an election in the database.
+                   Note that this does not allow you to change the _name_ of an election as this would generate a new slug."""
+)
```
EarthenSky, line 122:
```suggestion
@router.patch(
    "/election/{slug:str}",
    description="""Updates an election in the database.
                   Note that this does not allow you to change the _name_ of an election as this would generate a new slug."""
)
```

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Default start time should be now unless specified otherwise
+    if start_datetime is None:
+        start_datetime = datetime.now()
+
+    if election_type not in election_types:
+        raise RequestValidationError()
+
+    params = ElectionParameters(
+        _slugify(name),
```
EarthenSky, line 80: since `slug` is used as a unique identifier, what if we have a duplicate slug here ("My election" vs "my election", etc...)? We should have a check to make sure that slug does not currently exist, so we don't accidentally overwrite a past election.

src/elections/urls.py
```diff
@@ -0,0 +1,160 @@
...
+    session_id = request.cookies.get("session_id", None)
+    user_auth = await _validate_user(db_session, session_id)
+    if user_auth is False:
+        # let's workshop how we actually wanna handle this
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail="You do not have permission to access this resource",
+            headers={"WWW-Authenticate": "Basic"},
+        )
+    if slug is not None:
```
EarthenSky, line 142: slug is not allowed to be none here, since fastapi actually enforces input types (iirc) -> it's worth a simple test anyways as a santity check

src/main.py
```diff
@@ -1,11 +1,16 @@
...
+from fastapi.encoders import jsonable_encoder
+from fastapi.exceptions import RequestValidationError
+from fastapi.responses import JSONResponse
 
 import auth.urls
 import database
+import elections.urls
 import officers.urls
 import permission.urls
+import tests.urls
```
EarthenSky, line 13: oops! This is from the merge of main into this branch; please remove this (was something weird I was doing, give it no mind)
```suggestion
```

src/permission/types.py
```diff
@@ -29,6 +31,21 @@ async def has_permission(db_session: database.DBSession, computing_id: str) -> b
 
         return False
 
+class ElectionOfficer:
+    @staticmethod
+    async def has_permission(db_session: database.DBSession, computing_id: str) -> bool:
+        """
+        An current elections officer has access to all elections, prior elections officers have no access.
+        """
+        officer_terms = await officers.crud.current_executive_team(db_session, True)
```
EarthenSky, line 40: `current_executive_team` has been replaced with `current_officers`. The behaviour *shooooould* be the same

src/permission/types.py
```diff
@@ -29,6 +31,21 @@ async def has_permission(db_session: database.DBSession, computing_id: str) -> b
...
+        current_election_officer = officer_terms.get(officers.constants.OfficerPosition.ElectionsOfficer.value)[0]
```
EarthenSky, line 41: what if there are multiple active elections officers? I can imagine this scenario. Instead of using the function `current_officers`, instead get the list of all active officer positions a user has & check if one of them is elections officer. This can be done more easily using `get_officer_terms`

EarthenSky (commented):
Oops, since I made this PR originally I can't request changes. However, consider this "request changes".

The biggest issue is the uniqueness of the slug, the rest are medium level.

Mostly the logic is good!

src/elections/crud.py
EarthenSky: that is true! i wasn't suggesting to commit here either, just talking about the hypothetical behaviour. I suppose even if you don't commit, doing `db_session.execute(...)` emits SQL statements. 

Looks at this stackoverflow https://stackoverflow.com/questions/9667138/how-to-update-sqlalchemy-row-entry#:~:text=UPDATE%20using%20SQLAlchemy%3A-,user.no_of_logins%20%2B%3D%201,-session.commit()

it implies that setting `election.start_datetime = params.start_datetime` actually emits some SQL statements! I don't like this choice by sqlalchemy, but there's not much we can do... I think this means you do not need the update statement at the end of this file. Please test it & get back to me if possible!

EarthenSky: OR, you can just remove the update at the end & say yolo

EarthenSky:
This PR creates tables for elections and nominees, while introducing tested endpoints just for the elections (C,R,U,D)

PR #67, opened 2024-08-17
Add admin log

EarthenSky:
Not sure if we're going to want this for the time being; maybe just logging all changes to a file is a better idea? Yeah...


jbriones1:
Closing this since it's stale

PR #78, opened 2024-08-24
Add foreign keys & update alembic revisions

EarthenSky:
I learned about databases and now understand the pure awesomeness of foreign keys.

Also cleaned up the alembic revisions (because I was scared of alembic and needed to not be) & added a naming convention to the metadata. Woo!


EarthenSky:
Tested & it seems to work correctly!

EarthenSky:
"Move fast & break things"

PR #92, opened 2024-11-16
Remove config files and create alembic workflow

micahdbak:
As per the creation of https://github.com/CSSS/csss-site-config, some configuration files that are directly related to the deployment to the server are now redundant.

We are aiming to make csss-site-backend/csss-site-frontend contain code irrespective of deployment environments, and have csss-site-config contain all configuration and deployment things moving forward, which will make handling multiple repositories a bit easier.

Also, new workflow to make sure that alembic upgrade head successfully happens for pull requests, to prevent broken db migrations.


EarthenSky (approved):
looks good, love it!

PR #94, opened 2024-12-09
Make `officers` module more DRY

### LINKED ISSUE #87: Cleanup officers module
It's messy and confuzzling, and needs to be made more dry

EarthenSky:
Close https://github.com/CSSS/csss-site-backend/issues/87

TODO:
- [x] implement and test the delete term endpoint
- [x] switch from datetime to date for dates
- [x] add all checks to OfficerInfoUpload.valid_or_raise & OfficerTermUpload.valid_or_raise
- [x] fix all TODOs in the module, or create issues for them
- [x] tests that all endpoints seem to work correctly (for GET)

see https://github.com/CSSS/csss-site-backend/issues/102 for final changes


micahdbak:
what's going on with the alembic upgrade head check? something with requirements? 😨

EarthenSky:
> alembic upgrade head check

I noticed it was failing but haven't looked into it yet. It's probably because I modified the existing alembic migration (changing from datetime -> date).

I updated the migration directly because I know we haven't deployed the database yet so it's easy to simply reset the db with the most up to date schema without losing any data. This is the last time we get to do that because I'm planning to add the officer data tonight \:)

micahdbak:
> I updated the migration directly because I know we haven't deployed the database yet so it's easy to simply reset the db with the most up to date schema without losing any data. This is the last time we get to do that because I'm planning to add the officer data tonight :)

Hmm but the test runs on an empty db and fresh settings for alembic, which means it can't be created from scratch

Regardless I'd like to get this merged ASAP to make sure my FE works with the routes; I'm working on the election *term* updating feature RN, then next will be an admin page for adding new officers + terms

micahdbak:
BTW, when you use the get officer info route, it returns discord ID, name, and username, but the patch route only takes name (as of what's in main); does ur PR fix that?

EarthenSky:
> Hmm but the test runs on an empty db ...

looks like it was due to the github action not our code; lucky!

> BTW, when you use the get officer info route, it returns discord ID, name, and username ...

responded in discord!

micahdbak (approved):
🔥

PR #112, opened 2025-08-26
Various fixes on Elections API

jbriones1:
* Added a response model for the elections and nominee endpoints
* Fixed some invalid HTTP 404 status code returns
* Fixed some return type issues on user validations functions

### DIFF
--- a/src/utils/urls.py
+++ b/src/utils/urls.py
@@ -1,6 +1,7 @@
 from fastapi import HTTPException, Request
 
 import auth
+import auth.crud
 import database
 
 # TODO: move other utils into this module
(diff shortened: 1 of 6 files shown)


src/utils/urls.py
p-north, line 1: This works as well. Although, it would just return an empty list if the user wasn't logged in. I guess the boolean avoids any ambiguity

src/utils/urls.py
```diff
@@ -23,7 +24,7 @@ async def logged_in_or_raise(
 async def is_logged_in(
     request: Request,
     db_session: database.DBSession
-) -> tuple[str | None, str | None]:
+) -> tuple[bool, str | None, str | None]:
```
p-north, line 27: This works as well. Although, it would just return an empty list if the user wasn't logged in. I guess the boolean avoids any ambiguity

PR #114, opened 2025-09-05
Refactor login flow

### LINKED ISSUE #35: Use search params for the redirect URL (auth)
Instead of providing the entire URL to redirect to (e.g., https://new.sfucsss.org/#profile), populate search params instead.

E.g.,

https://api.sfucsss.org/auth/login?page=profile -> https://new.sfucsss.org/#profile
https://api.sfucsss.org/auth/login?app=elections&page=2025/1 -> https://new.sfucsss.org/elections/#2025/1
...

### LINKED ISSUE #65: Only pass session id cookie to api
> Note that if a cookie's [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#samesitesamesite-value) attribute is set to Strict or Lax, then the cookie will not be sent cross-site, even if credentials is set to include.

We should look into only sending the session id cookie to the api endpoint.

Since our session tokens are reasonably sized (& intended to be private), it's a good idea to not go throwing them at every page & resource GET. There's no security or performance issue of course (latency is way slower than several bytes of data), but it's a good idea to only do what's necessary.

Unless of course, we have some need in the future for non-api pages to serve restricted content? (https://github.com/CSSS/csss-site-backend/issues/54)

### LINKED ISSUE #88: Update auth API routes with /api in mind
I.e. next route shouldn't require the host and method

Can have an optional "base" param which would be new.sfucsss.org, sfucsss.org, dev-branch.sfucsss.org, ...

### LINKED ISSUE #91: Update /auth/login to be POST instead of GET
I learned recently that all GET requests [must be safe, idemponent, and cacheable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET). Imagine if a GET request had some important side-effect, but a proxy cached the result, and returned it without informing the server! Then the side-effect would not occur.

From https://www.sfu.ca/information-systems/services/cas/cas-for-web-applications/
> When CAS redirects the authenticated user back to your application, it will append a parameter named ticket to your URL. For instance, the URL you supplied CAS may be called as follows:
> 
> http://my/url?ticket=ST-92834-m34Aa83f7a3f

Since the CAS auth system redirects the web-browser to an endpoint, the browser will only call GET on that endpoint. This means we need to supply both a GET and a POST endpoint which accomplishes logging in.

Context
> idemponent, meaning "the intended effect on the server of making a **single** request is the same as the effect of making **several** identical requests". This seems like it isn't the case, since making several requests will start reporting that the GET request failed. However, the internal effect of the system will be the same: logging in 3 times will still result in the user being logged in (or result in a failure 3 times).
> 
> However that's a really good point! We should add a comment that says why it's GET, because I didn't realize that GET / POST were enforced strictly until now either!
> 
> That's a good point; I'm worried that since the redirect redirects the browser to the login endpoint (so that the backend can authenticate the user given the service token from sfu), the browser will automatically throw a GET request at the endpoint.
> 
> Though you're totally right we should expect the same behaviour. I suppose the solution is to redirect to a frontend page which runs some JS to call the backend, but I don't like the idea of zealously relying on js. Something as simple as login should be possible from a browser w/ js off, since some people are going to be logging in as their first interaction with the website...

### LINKED ISSUE #113: Refactor login flow
## Description
Right now our backend login is heavily reliant on our frontend being `https://sfucsss.org`. I would like to make this origin agnostic, so that if we build any other application, we only need to rely on a single backend for authentication. This would allow us to have a single CSSS authentication session_id which can be shared between any other application.

Our current login flow looks like this:
1. Frontend, on page `https://sfucsss.org/some-page` (must be sfucsss.org, doesn't work with other URLs)
    1.1. Navigate to `https://cas.sfu.ca/cas/login?service=https://sfucsss.org/api/auth/login?redirect_path=some-page&redirect_fragment=`  (service value is URI encoded)
2. SFU CAS
    2.1. Log in with computing ID and password
    2.2. Get redirected to `https://sfucsss.org/api/auth/login?redirect_path=some-page&redirect_fragment=`
    2.3. On the `new` frontend, this triggers the endpoint `/api/auth/login?redirect_path=some-page&redirect_fragment=`
3. Backend, on route `/api/auth/login`
    3.1. Contact the SFU CAS validation system with the service and the ticket number provided by the frontend
    3.2a. If the service and ticket matched, create a session cookie, and then a redirect (HTTP 307) is sent back with `location` header going to `https://sfucsss.org/some-page#`
    3.2b. If the validation with SFU's CAS fails, then send back a 401

This whole login flow assumes that:
* We are on `https://sfucsss.org` (or http://localhost:8080)
* The application requires us to be routed to a specific page
* We require a fragment, since the new site used hash routing for the SPA application

## Changes
* Make login a POST request (see #91), with body
```python
class LoginBodyModel(BaseModel)
    service: str # The URL used for the initial CAS ticket request
    ticket: str # The ticket number sent back by CAS
    redirect_url: str | None # Optional route that the website should redirect to. Mainly for sites that don't want to handle redirects themselves. Always appended to the origin location so we limit where people are redirected to
```
* Add [CORS middleware](https://fastapi.tiangolo.com/tutorial/cors/) to allow several different allowed origins
* Update the cookie sent to only allow subdomains we allow

The login flow would look like this:
1. Frontend, on page URL `https://sfucsss.org/auth` (or `https://admin.sfucsss.org/auth`)
    1.1. Navigate to `https://cas.sfu.ca/cas/login?service=https://sfucsss.org/auth` (service value is URI encoded)
2. SFU CAS
    2.1. Log in with computing ID and password
    2.2. Get redirected to `https://sfucsss.org/auth`
    2.3. Frontend *manually* calls endpoint `https://sfucsss.org/api/auth/login` with appropriate body
3. Backend, on route `/api/auth/login`
    3.1. Contact the SFU CAS validation system with the service and the ticket number provided by the frontend
    3.2a. If the service and ticket match, create a session cookie and if there is a redirect URL, then a redirect (HTTP 307) is sent back with `location` header going to `https://<origin>/<redirect_url>`. Otherwise, send back a 200, which is handled by the frontend however it wants
    3.2b. If the validation with SFU's CAS fails, then send back a 401

This new login flow allows us to use our authentication API with multiple frontends and allow the frontend to decide what it wants to do if it has been authenticated.

jbriones1:
* closes #88: CORS and redirects using the `origin` requires the host be supplied
* closes #35: Redirect URL/URI should be supplied in full, will look into fragments later
* closes #65: Always send the cookie, since we restrict our API to only certain origins and we currently only ask the backend for API requests since we use a bunch of SPA apps
* closes #91: Request parameters are now stored in the body.
* closes #113 (main focus): Refactored the login flow, to match what was explained in the issue

### DIFF
--- a/src/auth/urls.py
+++ b/src/auth/urls.py
@@ -2,11 +2,10 @@ import base64
 import logging
 import os
 import urllib.parse
-from typing import Literal
 
 import requests  # TODO: make this async
 import xmltodict
-from fastapi import APIRouter, BackgroundTasks, Body, HTTPException, Request, Response
+from fastapi import APIRouter, BackgroundTasks,  HTTPException, Request, Response
 from fastapi.responses import JSONResponse, PlainTextResponse, RedirectResponse
 
 import database
--- a/src/auth/urls.py
+++ b/src/auth/urls.py
@@ -2,15 +2,18 @@ import base64
 import logging
 import os
 import urllib.parse
+from typing import Literal
 
 import requests  # TODO: make this async
 import xmltodict
-from fastapi import APIRouter, BackgroundTasks, HTTPException, Request
+from fastapi import APIRouter, BackgroundTasks, Body, HTTPException, Request, Response
 from fastapi.responses import JSONResponse, PlainTextResponse, RedirectResponse
 
 import database
 from auth import crud
-from constants import FRONTEND_ROOT_URL
+from auth.models import LoginBodyModel
+from constants import IS_PROD
+from utils.shared_models import DetailModel
 
 _logger = logging.getLogger(__name__)
 
@@ -32,27 +35,34 @@ router = APIRouter(
 )
 
 
-# NOTE: logging in a second time invaldiates the last session_id
-@router.get(
+# NOTE: logging in a second time invalidates the last session_id
+@router.post(
     "/login",
-    description="Login to the sfucsss.org. Must redirect to this endpoint from SFU's cas authentication service for correct parameters",
+    description="Create a login session.",
+    response_description="Successfully validated with SFU's CAS",
+    response_model=None,
+    responses={
+        307: { "description": "Successful validation, with redirect" },
+        400: { "description": "Origin is missing.", "model": DetailModel },
+        401: { "description": "Failed to validate ticket with SFU's CAS", "model": DetailModel }
+    },
+    operation_id="login",
 )
 async def login_user(
-    redirect_path: str,
-    redirect_fragment: str,
-    ticket: str,
+    request: Request,
     db_session: database.DBSession,
     background_tasks: BackgroundTasks,
+    body: LoginBodyModel
 ):
     # verify the ticket is valid
-    service = urllib.parse.quote(f"{FRONTEND_ROOT_URL}/api/auth/login?redirect_path={redirect_path}&redirect_fragment={redirect_fragment}")
-    service_validate_url = f"https://cas.sfu.ca/cas/serviceValidate?service={service}&ticket={ticket}"
+    service_url = body.service
+    service = urllib.parse.quote(service_url)
+    service_validate_url = f"https://cas.sfu.ca/cas/serviceValidate?service={service}&ticket={body.ticket}"
     cas_response = xmltodict.parse(requests.get(service_validate_url).text)
 
     if "cas:authenticationFailure" in cas_response["cas:serviceResponse"]:
         _logger.info(f"User failed to login, with response {cas_response}")
-        raise HTTPException(status_code=401, detail="authentication error, ticket likely invalid")
-
+        raise HTTPException(status_code=401, detail="authentication error")
     else:
         session_id = generate_session_id_b64(256)
         computing_id = cas_response["cas:serviceResponse"]["cas:authenticationSuccess"]["cas:user"]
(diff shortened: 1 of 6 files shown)


src/constants.py
```diff
@@ -2,8 +2,9 @@
 
 # TODO(future): replace new.sfucsss.org with sfucsss.org during migration
 # TODO(far-future): branch-specific root IP addresses (e.g., devbranch.sfucsss.org)
-FRONTEND_ROOT_URL = "http://localhost:8080" if os.environ.get("LOCAL") == "true" else "https://new.sfucsss.org"
+FRONTEND_ROOT_URL = "http://localhost:8080" if os.environ.get("LOCAL") == "true" else "https://sfucsss.org"
 GITHUB_ORG_NAME = "CSSS-Test-Organization" if os.environ.get("LOCAL") == "true" else "CSSS"
+IS_PROD = os.environ.get("LOCAL") == "false"
```
p-north, line 7: This will only return true if the string is literally "false".
I would suggest doing -> `IS_PROD = os.environ.get("LOCAL", "").lower() != "true"`

src/auth/urls.py
```diff
@@ -2,15 +2,18 @@
 import logging
 import os
 import urllib.parse
+from typing import Literal
```
p-north, line 5: Unused. can be removed

src/constants.py
p-north, line 7: This will only return true if the string is literally "false".
I would suggest doing -> `IS_PROD = os.environ.get("LOCAL", "").lower() != "true"`

src/constants.py
```diff
@@ -25,3 +26,7 @@
 
 # https://docs.github.com/en/enterprise-server@3.10/admin/identity-and-access-management/iam-configuration-reference/username-considerations-for-external-authentication
 GITHUB_USERNAME_LEN = 39
+
+# COOKIE
+SAMESITE=None if IS_PROD else "lax"
```
p-north, line 31: In FastAPI, **samesite** expects "strict" | "lax" | "none". 
I suggest changing to -> `SAMESITE= "none" if IS_PROD else "lax"`

src/constants.py
```diff
@@ -25,3 +26,7 @@
...
+DOMAIN=".sfucsss.org" if IS_PROD else None
```
p-north, line 32: Some browsers may reject cookies with **None** domain. Consider using something like `"localhost"` instead?

p-north (changes requested):
-Mainly looks good. Consider the suggestions and minor tweaks here and there.

src/constants.py
jbriones1: For localhost it's suggested to leave it out completely, which is what happens when you supply `None` to the `set_cookie()` function.
image

p-north: oki

p-north (approved):
-Looks Good!

PR #115, opened 2025-09-07
Update: Auth API

jbriones1:
* Switched to using the DeclarativeBase type from SQLAlchemy for better type checking with PyRight
* Updated the Auth tables (site_user, user_session) to use ORM mapped values, to remove type errors when comparing `Column[...]` to Python-typed values
* Added body and response models to the auth endpoints
* Added a bunch of typing

### DIFF
--- a/src/auth/tables.py
+++ b/src/auth/tables.py
@@ -1,7 +1,6 @@
 from datetime import datetime
-from typing import Any
 
-from sqlalchemy import DateTime, ForeignKey, String, Text
+from sqlalchemy import DateTime, ForeignKey, String, Text, func
 from sqlalchemy.orm import Mapped, mapped_column
 
 from constants import COMPUTING_ID_LEN, SESSION_ID_LEN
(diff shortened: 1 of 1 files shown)


src/auth/tables.py
```diff
@@ -29,15 +31,23 @@ class SiteUser(Base):
...
+    computing_id: Mapped[str] = mapped_column(
         String(COMPUTING_ID_LEN),
         primary_key=True,
     )
 
     # first and last time logged into the CSSS API
     # note: default date (for pre-existing columns) is June 16th, 2024
-    first_logged_in = Column(DateTime, nullable=False, default=datetime(2024, 6, 16))
-    last_logged_in = Column(DateTime, nullable=False, default=datetime(2024, 6, 16))
+    first_logged_in: Mapped[datetime] = mapped_column(DateTime, nullable=False, default=datetime(2024, 6, 16))
```
p-north, line 41: is this supposed to be hardcoded?
I think it would be better to do something like `default=datetime.utcnow`?

src/auth/tables.py
```diff
@@ -29,15 +31,23 @@ class SiteUser(Base):
...
+    last_logged_in: Mapped[datetime] = mapped_column(DateTime, nullable=False, default=datetime(2024, 6, 16))
```
p-north, line 42: same here

p-north (changes requested):
-Looks good, refer to sugestions below

src/auth/tables.py
jbriones1: The comment Gabe left I guess left it as June 16, 2024 to indicate when they took over the project? I agree it should be changed, so I'll make it the server default, according to the SQLAlchemy docs

https://docs.sqlalchemy.org/en/20/orm/declarative_tables.html#mapping-whole-column-declarations-to-python-types

p-north: oki

p-north (approved):
-looks good to me

PR #118, opened 2025-09-08
Update: Various Elections API changes, Part 2

jbriones1:
### Updates
* Updated the metadata for the elections endpoints so that our docs are accurate
* Singularized the election endpoints and moved the election endpoints to their own root path
    * `/elections` -> `/election`
    * `/elections/registration/` -> `/registration`
    * `/elections/nominee_info/` -> `/nominee`
    * This enforces a pattern where each table has its own set of operations. We can always make endpoints that do more complicated things, which compose multiple operations in a single call.
* Changed GET `/elections/list` -> GET `/election`
    * Removed the constraint that elections can't be named "list"
*  Updated all PATCH endpoints to the following pattern:
    * The URL params are used to target a specific database entry
    * The body holds the values that will be used to update the entry
    * The entry's fields will only be updated if there is an explicit value set in the body (so a `body.speech` that's `NULL` will set it to `NULL` in the database, but if `body.speech` doesn't exist then the entry's value will not change)
* Updated all the POST endpoints to use the body as the entry's values
* DELETE endpoints target a single entry in the database using the URL parameters
* Made some endpoints require you to be an admin
* All current elections unit tests should all pass
* Made enums for officer positions and election statuses
* `available_positions` will automatically be converted to a `str` when saved to the database and to a `list[str]` when used in the Python code
* Moved nominee-application-related things to the `registrations` folder
* Moved nominee-info-related things to the `nominees` folder

### Todo (Part 3)
* DELETE `/nominee/{election_name}/{computing_id}/{position`: to delete nominees
* POST `/nominee`: to create nominees (PATCH does this, but semantically that's kind of weird)
* Add unit tests for all the newer endpoints
* Break up unit tests for easier debugging
* Remove some constant strings that are floating around and convert them to enums if possible

### DIFF
--- a/src/utils/types.py
+++ b/src/utils/types.py
@@ -15,7 +15,7 @@ class StringList(TypeDecorator):
 
         return ",".join(value)
 
-    def process_result_value(self, value, dialect: Dialect) -> list[str] | None:
+    def process_result_value(self, value, dialect: Dialect) -> list[str]:
         if value is None or value == "":
             return []
         return value.split(",")


src/nominees/tables.py
```diff
@@ -0,0 +1,48 @@
...
+    computing_id: Mapped[str] = mapped_column(String(COMPUTING_ID_LEN), primary_key=True)
+    full_name: Mapped[str] = mapped_column(String(64), nullable=False)
+    linked_in: Mapped[str] = mapped_column(String(128))
+    instagram: Mapped[str] = mapped_column(String(128))
+    email: Mapped[str] = mapped_column(String(64))
+    discord_username: Mapped[str] = mapped_column(String(DISCORD_NICKNAME_LEN))
+
+    def to_update_dict(self) -> dict:
+        return {
+            "computing_id": self.computing_id,
```
p-north, line 25: **computing_id** is the primary key. Shouldn't be allowed to update this in dictionaries usually.

src/nominees/urls.py
```diff
@@ -0,0 +1,104 @@
...
+    # if not already existing, create it
+    if not existing_info:
+        # unpack dictionary and expand into NomineeInfo class
+        new_nominee_info = NomineeInfo(computing_id=computing_id, **updated_data)
+        # create a new nominee
+        await nominees.crud.create_nominee_info(db_session, new_nominee_info)
+    # else just update the partial data
+    else:
+        merged_data = {
+            "computing_id": computing_id,
```
p-north, line 84: Again, don't think the PK should be allowed to update.

src/nominees/urls.py
```diff
@@ -0,0 +1,104 @@
...
+    },
+    operation_id="update_nominee"
+)
+async def provide_nominee_info(
+    request: Request,
+    db_session: database.DBSession,
+    body: NomineeInfoUpdateParams,
+    computing_id: str
+):
+    # TODO: There needs to be a lot more validation here.
```
p-north, line 58: Could also check for string validation to ensure they don't reach the DB max.

src/registrations/models.py
```diff
@@ -0,0 +1,28 @@
+from pydantic import BaseModel
+
+from officers.constants import OfficerPositionEnum
+
+
+class RegistrationModel(BaseModel):
+    position: str
```
p-north, line 7: position is typed as **str**. Might want to use the position enum instead.

src/registrations/urls.py
```diff
@@ -0,0 +1,256 @@
...
+
+    if body.position not in election.available_positions:
+        # NOTE: We only restrict creating a registration for a position that doesn't exist,
+        # not updating or deleting one
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"{body.position} is not available to register for in this election"
+        )
+
+    if election.status(datetime.datetime.now(tz=datetime.UTC)) != ElectionStatusEnum.NOMINATIONS:
```
p-north, line 108: -Double check, but I don't think **datetime.UTC** is a valid attribute. Might be -> **datetime.timezone.utc.**

OR even -> **datetime.now(timezone.utc())**

src/registrations/urls.py
```diff
@@ -0,0 +1,256 @@
...
+)
+async def register_in_election(
+    request: Request,
+    db_session: database.DBSession,
+    body: NomineeApplicationParams,
+    election_name: str
+):
+    await admin_or_raise(request, db_session)
+
+    if body.position not in OfficerPositionEnum:
```
p-north, line 79: might have to do -> `if body.position is not None and body.position not in OfficerPositionEnum:`

src/utils/types.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def process_bind_param(self, value, dialect: Dialect) -> str:
+        if value is None:
+            return ""
+        if not isinstance(value, list):
+            raise ValueError("Must be a list")
+
+        return ",".join(value)
+
+    def process_result_value(self, value, dialect: Dialect) -> list[str] | None:
```
p-north, line 18: This never returns **None**. Both **None** and **""** are converted to an empty list [] either way so better to just keep type hint to only **list[str]**.

src/utils/types.py
```diff
@@ -0,0 +1,23 @@
...
+    impl = Text
+    cache_ok = True
+
+    def process_bind_param(self, value, dialect: Dialect) -> str:
+        if value is None:
+            return ""
+        if not isinstance(value, list):
+            raise ValueError("Must be a list")
+
+        return ",".join(value)
```
p-north, line 16: Can just use **json.dumps(value)** here. 
Refer to this discussion: [Stack Overflow Discussion](https://stackoverflow.com/questions/32911336/what-is-the-difference-between-json-dumps-and-json-load)

src/utils/types.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def process_result_value(self, value, dialect: Dialect) -> list[str] | None:
+        if value is None or value == "":
+            return []
+        return value.split(",")
```
p-north, line 21: Same here, can just use **json.loads(value)**.
Refer to this discussion: [Stack Overflow Discussion](https://stackoverflow.com/questions/32911336/what-is-the-difference-between-json-dumps-and-json-load)

src/utils/urls.py
```diff
@@ -1,10 +1,17 @@
...
 import auth
 import auth.crud
 import database
+from permission.types import ElectionOfficer, WebsiteAdmin
+
 
 # TODO: move other utils into this module
+def slugify(text: str) -> str:
+    """Creates a unique slug based on text passed in. Assumes non-unicode text."""
+    return re.sub(r"[\W_]+", "-", text.strip().replace("/", "").replace("&", ""))
```
p-north, line 14: -Just use slugify here.

p-north (changes requested):
-Mostly fine, just refer to the suggestions below.

src/elections/tables.py
```diff
@@ -60,7 +62,7 @@ def private_details(self, at_time: datetime) -> dict:
             "datetime_end_voting": self.datetime_end_voting.isoformat(),
 
             "status": self.status(at_time),
-            "available_positions": self.available_positions,
+            "available_positions": self._available_positions,
```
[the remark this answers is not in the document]
jbriones1: I think I fixed this with [a6b60c1](https://github.com/CSSS/csss-site-backend/pull/118/commits/a6b60c14d0863c3302f0c5f9fc9a429d742ca716)

src/elections/urls.py
```diff
@@ -1,65 +1,110 @@
...
         has_permission = await WebsiteAdmin.has_permission(db_session, computing_id)
 
     return has_permission, session_id, computing_id
 
-# elections ------------------------------------------------------------- #
+def _default_election_positions(election_type: ElectionTypeEnum) -> list[OfficerPositionEnum]:
+    if election_type == ElectionTypeEnum.GENERAL:
+        available_positions = GENERAL_ELECTION_POSITIONS
+    elif election_type == ElectionTypeEnum.BY_ELECTION:
+        available_positions = GENERAL_ELECTION_POSITIONS
```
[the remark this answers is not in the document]
jbriones1: By elections and general elections have the same positions, since by elections are done to fill any positions that are left vacant after.

src/nominees/crud.py
```diff
@@ -0,0 +1,32 @@
...
+        .select(NomineeInfo)
+        .where(NomineeInfo.computing_id == computing_id)
+    )
+
+async def create_nominee_info(
+    db_session: AsyncSession,
+    info: NomineeInfo,
+):
+    db_session.add(info)
+
```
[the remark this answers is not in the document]
jbriones1: This seems to be a common pattern throughout the entire application: none of the table CRUD functions call commit; they're all called at the end of the request function to ensure things can rollback if the operation fails later down the line.

src/nominees/urls.py
jbriones1: I agree. We might need to review what should be editable on a table. This means for NomineeApplications, we can't change the nominee_election or position since they're part of the primary key. We'd have to create + delete to change someone's position they're running for

src/nominees/urls.py
jbriones1: We should leverage the built-in validation functions that SQLAlchemy has using the `@validates` decorator or use Pydantic models https://docs.pydantic.dev/latest/concepts/models/.

This would be a big change, so it might not be appropriate for this PR, maybe in Part 3.

src/registrations/urls.py
jbriones1: image

Ruff complains https://docs.astral.sh/ruff/rules/datetime-timezone-utc/

src/utils/types.py
jbriones1: The value coming out of the database isn't a JSON object, just a string with commas separating the positions:
`president,vice-president`.

src/utils/urls.py
jbriones1: This is the function definition of slugify, or are you suggesting we use a library to do this?

src/elections/tables.py
```diff
@@ -99,89 +86,35 @@ def to_update_dict(self) -> dict:
             "name": self.name,
             "type": self.type,
 
-            "datetime_start_nominations": self.datetime_start_nominations,
-            "datetime_start_voting": self.datetime_start_voting,
-            "datetime_end_voting": self.datetime_end_voting,
+            "datetime_start_nominations": self.datetime_start_nominations.date(),
+            "datetime_start_voting": self.datetime_start_voting.date(),
+            "datetime_end_voting": self.datetime_end_voting.date(),
```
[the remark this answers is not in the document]
jbriones1: Fixed, thanks

src/elections/tables.py
```diff
@@ -37,16 +21,19 @@ class Election(Base):
...
-    datetime_start_nominations = Column(DateTime, nullable=False)
-    datetime_start_voting = Column(DateTime, nullable=False)
-    datetime_end_voting = Column(DateTime, nullable=False)
-
-    # a csv list of positions which must be elements of OfficerPosition
-    available_positions = Column(Text, nullable=False)
-    survey_link = Column(String(300))
+    slug: Mapped[str] = mapped_column(String(MAX_ELECTION_SLUG), primary_key=True)
+    name: Mapped[str] = mapped_column(String(MAX_ELECTION_NAME), nullable=False)
+    type: Mapped[str] = mapped_column(String(64), default="general_election")
```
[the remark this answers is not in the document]
jbriones1: Fixed, thanks

src/elections/urls.py
```diff
@@ -267,387 +279,85 @@ async def create_election(
...
         )
 
+    election.update_from_params(body)
+
+    # TODO: We might be able to just use a validation function from Pydantic or SQLAlchemy to check this
+    _raise_if_bad_election_data(
+        slugified_name,
+        election.type,
+        election.datetime_start_voting,
+        election.datetime_start_voting,
```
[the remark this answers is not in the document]
jbriones1: Good catch, thanks

src/elections/urls.py
```diff
@@ -143,94 +192,52 @@ async def get_election(
...
+        slugified_name,
+        body.type,
+        start_nominations,
+        start_voting,
+        end_voting,
+        available_positions
     )
 
-    is_valid_user, _, _ = await _validate_user(request, db_session)
+    is_valid_user, _, _ = await get_user_permissions(request, db_session)
```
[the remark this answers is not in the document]
jbriones1: Will probably do this in a Part 3 or something.

src/utils/urls.py
p-north: I meant the function but just a nit pick.

p-north (approved):
-Looks good

src/nominees/urls.py
p-north: Sure!

PR #121, opened 2025-09-29
Update: Officers API

### LINKED ISSUE #2: REST api for officers

### LINKED ISSUE #76: Officer profile information
Officers are going to want to update & view all their info. This info should be displayed on the profile page.

Officers should only be able to update their term info for the current term, but their officer info any time.

Only officers who have an officerinfo entry in the db can view their info.

Related to https://github.com/CSSS/csss-site-backend/issues/57

### LINKED ISSUE #100: Return whether deletion succeeded or not
Right now we don't report anything to the frontend if the deletion fails; we should return if a deletion does not succeed (for whatever reason).

jbriones1:
Tests
- added Webmaster test db entries for easier testing
- refactored the test fixtures so they can be re-used
- fixed some warnings that would become errors in the future
- this patch breaks the elections tests, but will be fixed afterwards
- broke down the Officer units tests into smaller tests

Officer
- closes #2: updated the officers API for basic CRUD usage
- updated the HTTP decorators to update the documentation
- added appropriate models

Officer Term
- closes #76: added a unique constraint between `computing_id`, `position`, and `start_date`
- closes #100: return `{ success: True }` if the delete doesn't fail
- updated the HTTP decorators to update the documentation
- added appropriate models

Elections
- the datetime fields should be timezone unaware now
- we may need to update old timestamps to this time

### DIFF
--- a/src/officers/urls.py
+++ b/src/officers/urls.py
@@ -106,7 +106,8 @@ async def all_officers(
     """,
     response_model=list[OfficerTermResponse],
     responses={
-        401: { "description": "not authorized to view private info", "model": DetailModel }
+        401: { "description": "not logged in", "model": DetailModel },
+        403: { "description": "not authorized to view private info", "model": DetailModel }
     },
     operation_id="get_officer_terms_by_id"
 )
@@ -201,7 +202,7 @@ async def create_officer_term(
     return JSONResponse({ "success": True })
 
 @router.patch(
-    "/info/{computing_id:str}",
+    "/info/{computing_id}",
     description="""
         After election, officer computing ids are input into our system.
         If you have been elected as a new officer, you may authenticate with SFU CAS,
@@ -237,7 +238,7 @@ async def update_officer_info(
     return JSONResponse(updated_officer_info.serializable_dict())
 
 @router.patch(
-    "/term/{term_id:int}",
+    "/term/{term_id}",
     description="Update the information for an Officer's term",
     response_model=OfficerTermResponse,
     responses={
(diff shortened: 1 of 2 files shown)


src/officers/urls.py
```diff
@@ -89,6 +104,11 @@ async def all_officers(
         Get term info for an executive. All term info is public for all past or active terms.
         Future terms can only be accessed by website admins.
     """,
+    response_model=list[OfficerTermResponse],
+    responses={
+        401: { "description": "not authorized to view private info", "model": DetailModel }
+    },
+    operation_id="get_officer_terms_by_id"
```
p-north: 401 is for unauthenticated users. Should be 403 (Forbidden)

src/officers/urls.py
```diff
@@ -174,127 +198,104 @@ async def new_officer_term(
         ))
 
     await db_session.commit()
-    return PlainTextResponse("ok")
+    return JSONResponse({ "success": True })
 
 @router.patch(
-    "/info/{computing_id}",
+    "/info/{computing_id:str}",
```
p-north: I don't think FastAPI does type annotations for route params like `:str`.

src/officers/urls.py
```diff
@@ -174,127 +198,104 @@ async def new_officer_term(
...
-        "officer_info": updated_officer_info.serializable_dict(),
-        "validation_failures": validation_failures,
-    })
+    updated_officer_info = await officers.crud.get_new_officer_info_or_raise(db_session, computing_id)
+    return JSONResponse(updated_officer_info.serializable_dict())
 
 @router.patch(
-    "/term/{term_id}",
-    description=""
+    "/term/{term_id:int}",
```
p-north: Same here, i don't think FastAPI does params type annotations.

p-north (changes requested):
-lgtm just review the suggested nitpicks.

p-north (approved):
-looks good to me

PR #122, opened 2025-12-17
chore: update project configuration

jbriones1:
### Project
* Updated the README to indicate different ways to run the project
* Pinned the Python version to 3.11.X to ensure all pinned packages work
* Removed `requirements.txt`, all configuration is now in `pyproject.toml`

### uv
* Added `uv.lock`
* Updated `.gitignore` to ignore uv's build artifacts 

### Ruff
* Excluded `src/alembic`  from Ruff actions
* Ran the formatter over the entire project (minus alembic)
    * Note: This means that every change in `/src` should just be a formatting/linting update

### Pyright
* Added some different rules to the Pyright config


jbriones1:
included in #127

PR #127, opened 2026-01-02
Test: Nominee endpoint

jbriones1:
depends on #126

* Added basic integration tests for the Nominees endpoints
* TODO: Once the checks for election officers is completed, create the rest of the tests for election officers


src/nominees/urls.py
```diff
@@ -1,104 +1,111 @@
...
-async def get_nominee_info(
-    request: Request,
+async def get_all_nominees(
     db_session: database.DBSession,
-    computing_id: str
 ):
+    # Putting this behind a wall since there is private information here
+    nominees_list = await nominees.crud.get_all_nominees(db_session)
+
+    return JSONResponse([item.serialize() for item in nominees_list])
```
p-north: nitpick: but I think you can just do `return nominees_list` here.

src/officers/models.py
```diff
@@ -12,26 +16,105 @@
     "phone_number",
     "github_username",
     "google_drive_email",
-    "photo_url"
+    "photo_url",
 }
```
p-north: I know this was already here from before, but maybe if it's not being used, we can get rid of it?

src/officers/models.py
```diff
@@ -12,26 +16,105 @@
...
+            position=term.position,
+            start_date=term.start_date,
+            end_date=term.end_date,
+            biography=term.biography,
+        )
+
+    @computed_field
+    @property
+    def is_active(self) -> bool:
+        return is_active_term(start_date=self.start_date, end_date=self.end_date)
```
p-north: I'm okay with **computed_fields** however long term they might become an issue if we ever return large list.

src/scripts/migrate_from_about_officers.py
```diff
@@ -143,40 +145,37 @@ def get_key(officer):
...
+                discord_nickname=officer["discord_nickname"],
+                google_drive_email=officer["gmail"],
+                github_username=officer["github_username"],
             )
 
             existing_officer_info = await db_session.scalar(
-                sqlalchemy
-                .select(OfficerInfo)
-                .where(OfficerInfo.computing_id == new_officer_info.computing_id)
+                sqlalchemy.select(OfficerInfoDB).where(OfficerInfoDB.computing_id == new_officer_info.computing_id)
```
p-north: I think you might be missing the `sqlalchemy` import.

tests/integration/conftest.py
```diff
@@ -20,37 +20,39 @@ def suppress_sqlalchemy_logs():
...
-@pytest_asyncio.fixture(scope="session", loop_scope="session")
+
+@pytest_asyncio.fixture(scope="function", loop_scope="session")
+async def db_session(database_setup: DatabaseSessionManager):
+    async with database_setup.session() as session:
+        yield session
+
+
+@pytest_asyncio.fixture(scope="module", loop_scope="session")
 async def client() -> AsyncGenerator[Any, None]:
```
p-north: missing import for `AsyncGenerator, Any`

p-north (commented):
Sorry for the delay, just been trying to balance work and other things! Hopefully its better now. 

Everything looks good. Just some minor nitpicks here and there.

src/officers/models.py
jbriones1: We'd be losing everything from that field when you do the database migration.

src/officers/models.py
jbriones1: What's the alternative?

jbriones1 (commented):
Addressed some comments

src/officers/models.py
p-north: It's fine for now, I don't think the number of active members will ever exceed that high now that I think about it.

p-north (approved):
-lgtm

PR #142, opened 2026-05-14
fix: return 409 when deleting election with nominees

### LINKED ISSUE #131: Deleting an Election with Candidates fails
Due to Election slug being a foreign key in election_nominee_application the delete fails.

### Solution
Send back HTTP code 409 due to the foreign key constraint failing.

ruguoba:
## Problem

When attempting to delete an election that has associated nominees, the foreign key constraint (`election_nominee_application.election_slug` → `election.slug`) causes an `IntegrityError`. This results in an unhandled 500 Internal Server Error.

## Solution

Wrap the delete operation in a try-except block to catch `sqlalchemy.exc.IntegrityError`. On error:
1. Roll back the transaction
2. Return HTTP 409 (Conflict) with a clear error message explaining that nominees must be deleted first

## Changes

- Added `from sqlalchemy.exc import IntegrityError` import
- Wrapped `delete_election` + `commit` in try-except block
- Added 409 response to the endpoint's OpenAPI documentation

## Testing

1. Create an election
2. Add nominees to the election  
3. Try to delete the election → should return 409 with message
4. Delete the nominees first, then delete the election → should succeed

Fixes #131

### DIFF
--- a/src/elections/urls.py
+++ b/src/elections/urls.py
@@ -2,6 +2,7 @@ import datetime
 
 from fastapi import APIRouter, Depends, HTTPException, status
 from fastapi.responses import JSONResponse
+from sqlalchemy.exc import IntegrityError
 
 import candidates.crud
 import database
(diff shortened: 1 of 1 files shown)


jbriones1:
Hi, I was just wondering if you're a student at SFU.

ruguoba:
> 你好，请问你是不是西蒙弗雷泽大学的学生？

不是

jbriones1:
Sorry, PRs are only accepted from people from SFU or part of the W3 Committee

PR #145, opened 2026-05-14
Fix 131 election delete fk conflict 2

### LINKED ISSUE #131: Deleting an Election with Candidates fails

kathryntanardy:
Fix: Return 409 on election delete conflict
closes #131 

Issue: Deleting an election with linked nominee applications failed.

Cause: election_nominee_application references election.slug via FK.

Fix: Checked whether a nominee application depends on the election. If yes, raise an 409 error.

### DIFF
--- a/src/elections/urls.py
+++ b/src/elections/urls.py
@@ -289,12 +289,12 @@ async def delete_election(db_session: database.DBSession, election_name: str):
     try:
         await elections.crud.delete_election(db_session, slugified_name)
         await db_session.commit()
-    except IntegrityError:
+    except IntegrityError as err:
         await db_session.rollback()
         raise HTTPException(
             status_code=status.HTTP_409_CONFLICT,
             detail=f"{election_name} is still referenced by nominee applications.",
-        )
+        ) from err
 
     old_election = await elections.crud.get_election(db_session, slugified_name)
     return JSONResponse({"success": old_election is None})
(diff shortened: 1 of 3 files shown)


src/elections/urls.py
```diff
@@ -275,13 +275,23 @@ async def update_election(
...
 async def delete_election(db_session: database.DBSession, election_name: str):
     slugified_name = slugify(election_name)
 
+    existing_applications = await candidates.crud.get_all_candidates_in_election(db_session, slugified_name)
+    if existing_applications:
+        raise HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail=f"{election_name} is still referenced by nominee applications.",
+        )
+
```
jbriones1: I think instead of making another database call we can utilize the exception that would get thrown from trying to delete the election.  SQLAlchemy throws an `IntegrityError` when there's a foreign key constraint issue, iirc. 

I think you could wrap lines 295 and 296 in a try-except block and raise the 409 in the exception block. This ensures we don't hit the database twice.

jbriones1 (changes requested):

jbriones1 (approved):
Looks good, thank you for the changes!

PR #146, opened 2026-05-16
Feature/event issue #108

### LINKED ISSUE #108: [Request] Events endpoint
## Description
I'd like to store Event information somewhere and have our frontend be able to fetch it. I think there would need to be a new table created, as well as a new API endpoint.

The use-case for this is the frontend using it to build a calendar on the main site. The only thing the frontend needs for this are GET requests, but I may want to make something to make it easier to add events to the calendar.

If this can be done with Google Calendar or something, that's fine too, but show me a proof-of-concept so I don't have to research how to implement it.

### DB Table: `Event`
* `eid` (INTEGER): Event ID (primary key)
* `name` (VARCHAR): Name of the event
* `startTime` (DATETIME): Start time of the event
* `endTime` (DATETIME): End time of the event

*Maybes*
* `description`: Short description of the event. 
* `repeat` (VARCHAR/INTEGER): How often an event repeats
    * This one might be confusing, but either we use integers as agreed upon enum values (0 = never, 1 = bi-weekly,  2 = monthly, etc.) or we just write it down as a string. Events that happen yearly, but don't have a predictable date, like Frosh, Tech Fair, etc., should have a `0/"never"` for their value. This is so if we have a recurring event (like the bi-weekly events), we don't need to store each event; we just always receive these events and let the requester handle it however they want.

### REST API Endpoints
* All requests should return `200`or `404` unless stated otherwise.
* Frontend will send data as `application/x-www-form-urlencoded`
*Note: Question marks indicate something that may not need to be added*
* GET  `/event`: Get all events (might not be needed, but I'll put it here for completeness)
* GET  `/event/<year>`:  Get events that start OR end in this year
* GET  `/event/<year>/<month>`: Get the events that start OR end in this month and year
```jsonc
// GET response
[
    {
        "name": "string",
        "startTime": "string/date",
        "endTime": "string/date",
        "description?": "string",
        "repeat?": "string/number"
    },
// the rest
]
``` 
* POST  `/event`
    * `201`: successful creation
```jsonc
// POST body
{
    "name": "string",
    "startTime": "string/date",
    "endTime": "string/date",
    "description?": "string",
    "repeat?": "string/number"
}

// POST response
{
    "eid": "number",
    "name": "string",
    "startTime": "string/date",
    "endTime": "string/date",
    "description?": "string",
    "repeat?": "string/number"
}
```
* PATCH `/event/<eid>`
```jsonc
// PATCH body
{
    "name": "string",
    "startTime": "string/date",
    "endTime": "string/date",
    "description?": "string",
    "repeat?": "string/number"
}


// PATCH response
{
    "eid": "number",
    "name": "string",
    "startTime": "string/date",
    "endTime": "string/date",
    "description?": "string",
    "repeat?": "string/number"
}
```
* DELETE `/event/<eid>`
    * `404`: If nothing was deleted
```jsonc
// DELETE body
{
    "eid": "number"
}

// DELETE response
{
    "result": "success | failure",
    "eid": "number" 
}
```

arunPdl02:
This PR implements the event feature that allows CRUD operations.

Created database schema using sqlalchemy with proper constraints.

Implement the following APIs:
GET( /event)
GET( /event/<year>)
GET(/event/<year>/<month>)
POST (/event)
PATCH (/event/<eid>)
DELETE (/event/<eid>)

Closes #108


src/alembic/versions/f4c493a24799_create_event_table.py
```diff
@@ -0,0 +1,42 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('event_info',
+    sa.Column('eid', sa.Integer(), autoincrement=True, nullable=False),
+    sa.Column('description', sa.Text(), nullable=True),
+    sa.Column('name', sa.String(length=64), nullable=False),
+    sa.Column('start_time', sa.DateTime(timezone=True), nullable=False),
+    sa.Column('end_time', sa.DateTime(timezone=True), nullable=False),
+    sa.Column('repeat', sa.String(length=64), nullable=False),
+    sa.Column('start_date', sa.Date(), nullable=True),
+    sa.Column('end_date', sa.Date(), nullable=True),
```
jbriones1: Change these to `repeat_start/end_date` so its more clear what is being stored. Sorry, it wasn't clear in the spec.

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+class Event(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    eid: int
+    name: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    description: str | None = None
+    repeat: str | None = None
+    start_date: datetime.date | None = None
+    end_date: datetime.date | None = None
```
jbriones1: Update these to say `repeat_start/end_date`.

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+
+class EventPublic(BaseModel):
+    model_config = ConfigDict(from_attributes=True)
+    name: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    description: str | None = None
+    repeat: str | None = None
+    start_date: datetime.date | None = None
+    end_date: datetime.date | None = None
```
jbriones1: Update these to say `repeat_start/end_date`.

src/event/models.py
jbriones1, line 1: To reduce the amount of duplicated fields you can use inheritance. Makes it less error-prone to update the exact same field across multiple models.

```python
class EventPublic(BaseModel):
    name: str
    start_time:
   ...

class Event(EventPublic):
    eid: int
   # has all of the fields from EventPublic
```

Also, I think it's fine to return the `eid`, so you can remove `EventPublic` and just return `Event`.

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+
+class EventCreate(BaseModel):
+    name: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    description: str | None = None
+    repeat: str | None = None
+    start_date: datetime.date | None = None
+    end_date: datetime.date | None = None
```
jbriones1: Update these to say `repeat_start/end_date`.

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+        return self
+
+class EventUpdate(BaseModel):
+    name: str | None = None
+    start_time: datetime.datetime | None = None
+    end_time: datetime.datetime | None = None
+    description: str | None = None
+    repeat: str | None = None
+    start_date: datetime.date | None = None
+    end_date: datetime.date | None = None
```
jbriones1: Update these to say `repeat_start/end_date`.

src/event/tables.py
```diff
@@ -0,0 +1,70 @@
...
+        String(64)
+    )
+    start_date: Mapped[date] = mapped_column(
+        Date,
+        nullable=True
+    )
+    end_date: Mapped[date] = mapped_column(
+        Date,
+        nullable=True
+    )
```
jbriones1: Update these to say `repeat_start/end_date`.

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+router = APIRouter(
+    prefix="/event",
+    tags=["event"],
+)
+
+@router.get(
+    "",
+    description="Get all events",
+    response_model=list[EventPublic],
+    # responses={},
```
jbriones1: Unnecessary comment

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+    operation_id="get_all_events",
+    # probably want it to be public so no dependecies?
+    # dependecies=[Depends()]
```
jbriones1: I agree, public, so no dependencies.

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+    events_list = await event.crud.get_all_events(db_session)
+
+    return events_list
+
+
+@router.get(
+    "/{year}",
+    description="Get events that start OR end in this year",
+    response_model=list[EventPublic],
+    # responses= {}
```
jbriones1: Unnecessary comment

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+    events_list = await event.crud.get_events_for_this_year(db_session, year)
+
+    return events_list
+
+
+@router.get(
+    "/{year}/{month}",
+    description="Get events that start OR end in the given year and month",
+    response_model=list[EventPublic],
+    # responses= {}
```
jbriones1: Unnecessary comment

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+@router.post(
+    "",
+    description="Create a new event",
+    response_model=Event,
+    status_code=status.HTTP_201_CREATED,
+    responses={
+        500: {"description": "failed to fetch new event", "model": DetailModel},
+    },
+    operation_id="create_event",
+    # dependecies=[Depends()]
```
jbriones1: For now make the dependency for site admins only

src/event/tables.py
```diff
@@ -0,0 +1,70 @@
...
+        return{
+            "eid": self.eid,
+            "name": self.name,
+            "description": self.description,
+            "start_time": self.start_time,
+            "end_time": self.end_time,
+            "start_date": self.start_date,
+            "end_date": self.end_date,
+        }
+
```
jbriones1: I'd prefer using Pydantic's `model_dump()` if this isn't used anywhere. If it isn't used just remove it so others don't make their own serialization objects.

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+@router.patch(
+    "/{eid}",
+    description="Update an Event detail",
+    response_model=Event,
+    responses={
+        404:{"description": "Event doesn't exist."}
+    },
+    operation_id="update_event"
```
jbriones1: Add a site admin dependency.

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+        if not db_event.end_date:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="The event start date and event end date must be initilized at the same time"
+            )
+        if body.start_date > db_event.end_date:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+                detail="The event start date must be before the event end date"
+            )
```
jbriones1: Validation should be handled by the model validators you wrote. You can merge the data into a single Pydantic `Event` object, which should automatically trigger the validator. For example:

```python
db_event = await event.crud.get_event_by_eid(db_session, eid)
...
db_data = Event.model_validate(db_event).model_dump()
patch_data = body.model_dump(exclude_unset=True)
merged_data = {**db_data, **patch_data}
try:
    updated_data = Event(**merged_data) # validators would run here
except ValidationError as e:
    raise HTTPException(...)
```

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @model_validator(mode="after")
+    def validate_time_range(self) -> "EventUpdate":
+        if self.start_time and self.end_time:
+            if self.start_time > self.end_time:
+                raise ValueError("The event start time must be before end time")
+        return self
```
jbriones1: nit: You don't have to do this, but maybe you could add a separate validator in case this has a `repeat_start_date` and `repeat_end_date` on the object.

src/event/urls.py
```diff
@@ -0,0 +1,186 @@
...
+
+@router.delete(
+    "/{eid}",
+    description="Delete an event",
+    response_model=EventDelete,
+    responses={
+        404:{"description": "Event doesn't exist."}
+    },
+    operation_id="delete_event",
+    # dependecies=[Depends()],
```
jbriones1: Add a site admin dependency.

jbriones1 (changes requested):
Thanks for the PR, very clean and understandable code.

For future reference, you can just assume anything that creates, updates, or deletes things from the database requires site admin access. Reads might be harder to understand, so you can ask for those.

src/event/models.py
```diff
@@ -0,0 +1,69 @@
...
+    eid: int
+    name: str
+    start_time: datetime.datetime
+    end_time: datetime.datetime
+    description: str | None = None
+    repeat: str | None = None
+    start_date: datetime.date | None = None
+    end_date: datetime.date | None = None
+
+class EventPublic(BaseModel):
```
jbriones1: Returning the `eid` is fine here, so you can remove this model and just use `Event`.

jbriones1 (changes requested):

src/event/urls.py
arunPdl02: I refactored model.py and created a BaseEvent model that has all the validation logic and rest of the models inherit from it that was it's more maintainable

arunPdl02:
Hi Jon, I worked on all the suggestions you had, I did the following:
* Rename start_data and end_date to repeat_start_date and repeat_end_date. Revereted the alembic migration and created a new one.
* Removed EventPublic pydantic model and used Event for all responses (now they all include eid)
* Added the site_admin dependency to all create, update and delete APIs
* Refactored patch(/event) and model.py to move all the validation logic to the models. To do this, I created a BaseEvent model and other models just inherit from this model that way the validation logic is more streamlined.

src/event/tables.py
```diff
@@ -1,70 +1,24 @@
...
-    end_date: Mapped[date] = mapped_column(
-        Date,
-        nullable=True
-    )
+    eid: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    description: Mapped[str] = mapped_column(Text, nullable=True)
+    name: Mapped[str] = mapped_column(String(64))
+    start_time: Mapped[datetime] = mapped_column(DateTime(timezone=True))
+    end_time: Mapped[datetime] = mapped_column(DateTime(timezone=True))
+    repeat: Mapped[str] = mapped_column(String(64))
```
jbriones1: Just wondering, what will you be using the state how often something is repeating If they're strings, make a `StrEnum` with the possible values so that people have an idea of what they can use. You could also just use regular `Enum` so that they're smaller to store.

jbriones1 (changes requested):

src/event/tables.py
arunPdl02: I will update it to use Enum and use SQLAlchemy's Enum for the db which looks like it stores it as VARCHAR in postgres. Something like:
DAILY = "DAILY"
WEEKLY = "WEEKLY"
MONTHLY = "MONTHLY"
SEMESTERLY = "SEMESTERLY"
YEARLY = "YEARLY"

arunPdl02: Hi Jon, I hope you had a wonderful weekend. I completed this work and I ended up using StrEnum, it saved me for conversion overheads. I tried to use Enum type directly in the db but that didn't workout as planned so I ended up using VARCHAR + check constraint for it instead.

src/alembic/versions/4928dc3f0b07_create_event_table.py
```diff
@@ -26,9 +26,10 @@ def upgrade() -> None:
     sa.Column('name', sa.String(length=64), nullable=False),
     sa.Column('start_time', sa.DateTime(timezone=True), nullable=False),
     sa.Column('end_time', sa.DateTime(timezone=True), nullable=False),
-    sa.Column('repeat', sa.String(length=64), nullable=False),
+    sa.Column('frequency', sa.String(length=64), server_default=sa.text("'NONE'"), nullable=True),
```
jbriones1: What's the difference between NONE and a null value? If there is none maybe we can just use null to represent no repeat

jbriones1 (commented):
Just one last little thing. I'm sorry it took me a while to get back to you

src/alembic/versions/4928dc3f0b07_create_event_table.py
arunPdl02: NONE is used to represent non-repeating events instead of just leaving it as the null value which would be more ambiguous. Explicitly stating NONE, makes things clear.

jbriones1: If that's the case then set `nullable=False` for the column.

jbriones1 (changes requested):
Just one tiny nitpick and it should be good to go!

src/alembic/versions/4928dc3f0b07_create_event_table.py
arunPdl02: yes, that makes a lot of sense. All done!

PR #147, opened 2026-05-27
Fix 132 nominee fk conflict

### LINKED ISSUE #132: Deleting a Nominee with a Candidate fails
Due to Computing ID being a foreign key in election_nominee_application the delete fails.

kathryntanardy:
Close #132 

Issue: Deleting an election failed when nominees still had applications for it.

Cause: Due to Computing ID being a foreign key in election_nominee_application the delete fails.

Fix: Catch `IntegrityError` on delete and return **409 Conflict** when dependent applications exist.


jbriones1 (approved):
Looks good, thank you!

PR #148, opened 2026-06-02
Feat 144 grouped election endpoints

### LINKED ISSUE #144: Grouped election endpoints
## Description
Modify the `/election` GET endpoints to optionally return an array of nominees for each election. have an array of nominees attached to each election response object.

Use a query parameter of `with_nominees=<truthy>` to indicate that the response should contain the list of nominees for that election. Any other value or a missing value should not contain the list. See the link below for which values are considered truthy.

## Features
* `/election` updated
* `/election/{election_name}` updated
* Tests that validate this behaviour

## Resources
Docs on how query parameters work in FastAPI: https://fastapi.tiangolo.com/tutorial/query-params/#query-parameter-type-conversion

VerrillAng:
Feature: Grouped Election Endpoints
closes #144 

Description:
Adds optional nominees to `GET /election` and `GET /election/{name}` via _with_nominees=true_. 
Guests sees:
1. name
2. position
3. speech.

Admins also get:
1. computing_id
2. linked_in
3. instagram
4. email
5. discord_username

Features:
* Added `with_nominees` query param on list and single election GET endpoints (default false)
* Integration tests in test_elections.py for guest/admin list and single endpoints (with and without with_nominees), checking response shape and that private fields are hidden from guests

### DIFF
--- a/src/elections/crud.py
+++ b/src/elections/crud.py
@@ -1,12 +1,11 @@
 from collections.abc import Sequence
+from datetime import datetime
 
 import sqlalchemy
 from sqlalchemy.ext.asyncio import AsyncSession
 
-import candidates.crud
-import nominees.crud
 from candidates.tables import CandidateDB
-from elections.models import ElectionNomineeSummary
+from elections.models import ElectionNomineeSummary, ElectionResponse
 from elections.tables import ElectionDB
 from nominees.tables import NomineeInfoDB
 
@@ -44,41 +43,105 @@ async def delete_election(db_session: AsyncSession, slug: str) -> None:
     await db_session.execute(sqlalchemy.delete(ElectionDB).where(ElectionDB.slug == slug))
 
 
-async def get_all_nominees_by_election(
+async def get_all_elections_with_nominees(
     db_session: AsyncSession,
+    at_time: datetime,
     has_permission: bool,
-) -> dict[str, list[ElectionNomineeSummary]]:
+) -> list[ElectionResponse]:
     """
-    Fetches all nominees across all elections in a JOIN query.
-    Returns a dict mapping election slug -> list of ElectionNomineeSummary.
-    Only fetches contact fields (computing_id, linked_in, etc.) when has_permission is True.
+    Fetches all elections and their nominees in a single query.
+    Returns a list of ElectionResponse with candidates embedded.
     """
-    if has_permission:
-        query = sqlalchemy.select(
-            CandidateDB.nominee_election,
-            CandidateDB.position,
-            CandidateDB.speech,
-            NomineeInfoDB.full_name,
-            NomineeInfoDB.computing_id,
-            NomineeInfoDB.linked_in,
-            NomineeInfoDB.instagram,
-            NomineeInfoDB.email,
-            NomineeInfoDB.discord_username,
-        ).join(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
-    else:
-        query = sqlalchemy.select(
-            CandidateDB.nominee_election,
-            CandidateDB.position,
-            CandidateDB.speech,
-            NomineeInfoDB.full_name,
-        ).join(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
+    query = (
+        sqlalchemy.select(ElectionDB, CandidateDB, NomineeInfoDB)
+        .outerjoin(CandidateDB, ElectionDB.slug == CandidateDB.nominee_election)
+        .outerjoin(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
+    )
 
     rows = (await db_session.execute(query)).all()
 
-    nominees_by_election: dict[str, list[ElectionNomineeSummary]] = {}
-    for row in rows:
+    elections_by_slug = {}
+    for election, candidate, nominee_info in rows:
+        if election.slug not in elections_by_slug:
+            elections_by_slug[election.slug] = ElectionResponse(
+                slug=election.slug,
+                name=election.name,
+                type=election.type,
+                datetime_start_nominations=election.datetime_start_nominations,
+                datetime_start_voting=election.datetime_start_voting,
+                datetime_end_voting=election.datetime_end_voting,
+                available_positions=election.available_positions,
+                status=election.status(at_time),
+                survey_link=election.survey_link if has_permission else None,
+                candidates=[],
+            )
+
+        if (candidate is None) or (nominee_info is None) or (nominee_info.full_name is None):
+            continue
+
         if has_permission:
             nominee = ElectionNomineeSummary(
+                full_name=nominee_info.full_name,
+                position=candidate.position,
+                speech=candidate.speech or "No speech provided by this candidate",
+                computing_id=nominee_info.computing_id,
+                linked_in=nominee_info.linked_in,
+                instagram=nominee_info.instagram,
+                email=nominee_info.email,
+                discord_username=nominee_info.discord_username,
+            )
+        else:
+            nominee = ElectionNomineeSummary(
+                full_name=nominee_info.full_name,
+                position=candidate.position,
+                speech=candidate.speech or "No speech provided by this candidate",
+            )
+        elections_by_slug[election.slug].candidates.append(nominee)
+
+    return list(elections_by_slug.values())
+
+
+async def get_election_nominees(
+    db_session: AsyncSession,
+    election_slug: str,
+    has_permission: bool,
+) -> list[ElectionNomineeSummary]:
+    """
+    Fetches all nominees for a single election in one JOIN query.
+    Only fetches contact fields when has_permission is True.
+    """
+    if has_permission:
+        query = (
+            sqlalchemy.select(
+                CandidateDB.position,
+                CandidateDB.speech,
+                NomineeInfoDB.full_name,
+                NomineeInfoDB.computing_id,
+                NomineeInfoDB.linked_in,
+                NomineeInfoDB.instagram,
+                NomineeInfoDB.email,
+                NomineeInfoDB.discord_username,
+            )
+            .outerjoin(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
+            .where(CandidateDB.nominee_election == election_slug)
+        )
+    else:
+        query = (
+            sqlalchemy.select(
+                CandidateDB.position,
+                CandidateDB.speech,
+                NomineeInfoDB.full_name,
+            )
+            .outerjoin(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
+            .where(CandidateDB.nominee_election == election_slug)
+        )
+
+    rows = (await db_session.execute(query)).all()
+
+    candidates_list = []
+    for row in rows:
+        if has_permission:
+            candidate_entry = ElectionNomineeSummary(
                 full_name=row.full_name,
                 position=row.position,
                 speech=row.speech or "No speech provided by this candidate",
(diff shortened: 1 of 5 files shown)


src/elections/urls.py
```diff
@@ -72,26 +72,76 @@ def _raise_if_bad_election_data(
...
+                "speech": ("No speech provided by this candidate" if nomination.speech is None else nomination.speech),
+            }
+        candidates_list.append(candidate_entry)
+    return candidates_list
+
+
 @router.get(
     "",
-    description="Returns a list of all election & their status",
+    description="Returns a list of all election, their status and nominees (if requested)",
```
jbriones1: Pluralize. "Return a list of all elections,..."

src/elections/urls.py
```diff
@@ -72,26 +72,76 @@ def _raise_if_bad_election_data(
...
+                election_data["candidates"] = await _get_election_nominees(db_session, election, has_permission)
+            election_metadata_list.append(election_data)
     else:
-        election_metadata_list = [election.public_details(current_time) for election in election_list]
+        for election in election_list:
+            election_data = election.public_details(current_time)
+            # Get the nominees for the election
+            if with_nominees:
+                election_data["candidates"] = await _get_election_nominees(db_session, election, has_permission)
+            election_metadata_list.append(election_data)
```
jbriones1: This will query the database once for each election, which is expensive. It'd be better to create a new CRUD operation that gets all the election and nominee information in one query (if it's requested). Something like:
SELECT *
FROM election
LEFT JOIN election_nominee_application
    ON election.slug=election_nominee_info.nominee_election
So either the query on line 123 is executed or the query I suggested is executed. Make sure to use LEFT JOIN so elections with no candidates still appear.


You could even go one step further and define the fields to get from the database in the query i.e. don't fetch private details if the user does not have permission, but you don't have to.
If you do decide to remove the private details at the query level you can add `response_model_exclude_none=True` in the decorator, under `response_model` on line 114 and that will automatically strip away all fields that have `None` as their value.

src/elections/urls.py
```diff
@@ -100,14 +150,19 @@ async def list_elections(
     "/{election_name}",
     description="""
     Retrieves the election data for an election by name.
-    Returns private details when the time is allowed.
-    If user is an admin or election officer, returns computing ids for each candidate as well.
+    Returns private details when user is admin or election officer.
```
jbriones1: You can remove this line, since the next one explains this as well.

src/elections/urls.py
```diff
@@ -72,26 +72,76 @@ def _raise_if_bad_election_data(
...
+                "discord_username": nominee_info.discord_username,
+            }
+        else:
+            candidate_entry = {
+                "full_name": nominee_info.full_name,
+                "position": nomination.position,
+                "speech": ("No speech provided by this candidate" if nomination.speech is None else nomination.speech),
+            }
+        candidates_list.append(candidate_entry)
+    return candidates_list
```
jbriones1: Move this to `elections/crud.py`.

src/elections/urls.py
```diff
@@ -72,26 +72,76 @@ def _raise_if_bad_election_data(
         )
 
 
+async def _get_election_nominees(
+    db_session: database.DBSession,
+    election_row: ElectionDB,
+    has_permission: bool,
+) -> list[dict]:
```
jbriones1: Change the return type to `list[ElectionNomineeSummary]` and modify `candidate_entry` to use the Pydantic model. 

`list[dict]` is too ambiguous as a type and could lead to bugs later.

src/elections/urls.py
```diff
@@ -117,41 +172,14 @@ async def get_election(db_session: database.DBSession, computing_id: SessionUser
...
-                "speech": ("No speech provided by this candidate" if nomination.speech is None else nomination.speech),
-            }
-            if has_permission:
-                candidate_entry["computing_id"] = nomination.computing_id
-            election_json["candidates"].append(candidate_entry)
-
-        # after the voting period starts, all election data becomes public
-        return JSONResponse(election_json)
+        if with_nominees:
+            election_json["candidates"] = await _get_election_nominees(db_session, election, has_permission)
```
jbriones1: You can move these statements outside the `if has_permission` block to prevent the duplicate on line 181-182.

tests/integration/test_elections.py
```diff
@@ -57,21 +57,60 @@ async def test_read_elections(db_session: DBSession):
...
+    assert len(election_2_response["candidates"]) >= 1
+    for candidate in election_2_response["candidates"]:
+        assert "full_name" in candidate
+        assert "position" in candidate
+        assert "speech" in candidate
+        assert "computing_id" not in candidate
+        assert "linked_in" not in candidate
+        assert "instagram" not in candidate
+        assert "email" not in candidate
+        assert "discord_username" not in candidate
```
jbriones1: Write a helper function for this. If the shape of the response changes in the future we would need to update it on line 105 as well.

Also, you don't need to test the fields that should always appear (full_name, position, speech).

tests/integration/test_elections.py
```diff
@@ -57,21 +57,60 @@ async def test_read_elections(db_session: DBSession):
...
```
jbriones1: See above comment.

tests/integration/test_elections.py
```diff
@@ -164,28 +203,59 @@ async def test__delete_candidate(client: AsyncClient):
...
+    assert response.status_code == 200
+    elections_list = {election["slug"]: election for election in response.json()}
+    election_2_response = elections_list[slugify(TEST_ELECTION_2)]
+    assert "survey_link" in election_2_response
+    assert "candidates" in election_2_response
+    assert len(election_2_response["candidates"]) >= 1
+    for candidate in election_2_response["candidates"]:
+        assert "full_name" in candidate
+        assert "position" in candidate
+        assert "speech" in candidate
```
jbriones1: You don't need to test the fields that should always appear.

tests/integration/test_elections.py
```diff
@@ -164,28 +203,59 @@ async def test__delete_candidate(client: AsyncClient):
...
+    assert response.status_code == 200
+    election_2_response = response.json()
+    assert election_2_response != {}
+    assert "survey_link" in election_2_response
+    assert "candidates" in election_2_response
+    assert len(election_2_response["candidates"]) >= 1
+    for candidate in election_2_response["candidates"]:
+        assert "full_name" in candidate
+        assert "position" in candidate
+        assert "speech" in candidate
```
jbriones1: See above comment.

jbriones1 (changes requested):
Thank you for the PR and for writing tests! Lmk if you have any questions about the changes requests.

src/elections/urls.py
VerrillAng: Changed to 
`description="Return a list of all elections, their statuses and nominees (if requested)",`

src/elections/urls.py
VerrillAng: Changed to ist[ElectionNomineeSummary] and:
`candidate_entry = ElectionNomineeSummary(...)`

src/elections/urls.py
VerrillAng: Changed both `GET /election` and `GET /election/{name}` to prevent duplicates

tests/integration/test_elections.py
VerrillAng: Created `assert_public_candidate_fields(candidate: dict)` and `assert_private_candidate_fields(candidate: dict)`

src/elections/urls.py
VerrillAng: Added `get_all_nominees_by_election` helper function in `elections/crud.py` and used on `GET /election`

src/elections/crud.py
```diff
@@ -37,3 +42,96 @@ async def delete_election(db_session: AsyncSession, slug: str) -> None:
...
+            CandidateDB.nominee_election,
+            CandidateDB.position,
+            CandidateDB.speech,
+            NomineeInfoDB.full_name,
+            NomineeInfoDB.computing_id,
+            NomineeInfoDB.linked_in,
+            NomineeInfoDB.instagram,
+            NomineeInfoDB.email,
+            NomineeInfoDB.discord_username,
+        ).join(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
```
jbriones1: Needs to be `outerjoin` or else elections without a candidate don't get fetched.

src/elections/crud.py
```diff
@@ -37,3 +42,96 @@ async def delete_election(db_session: AsyncSession, slug: str) -> None:
...
+    else:
+        query = sqlalchemy.select(
+            CandidateDB.nominee_election,
+            CandidateDB.position,
+            CandidateDB.speech,
+            NomineeInfoDB.full_name,
+        ).join(NomineeInfoDB, CandidateDB.computing_id == NomineeInfoDB.computing_id)
```
jbriones1: Need to be `outerjoin` or else elections without a candidate don't get fetched.

src/elections/crud.py
```diff
@@ -37,3 +42,96 @@ async def delete_election(db_session: AsyncSession, slug: str) -> None:
...
+    election_row: ElectionDB,
+    has_permission: bool,
+) -> list[ElectionNomineeSummary]:
+    candidates_list = []
+    all_nominations = await candidates.crud.get_all_candidates_in_election(db_session, election_row.slug)
+    if not all_nominations:
+        return []
+    for nomination in all_nominations:
+        # NOTE: if a nominee does not input their legal name, they are not considered a nominee
+        nominee_info = await nominees.crud.get_nominee_info(db_session, nomination.computing_id)
```
jbriones1: I think this could also be changed so that you can batch all the requests into one DB query instead of having to hit the database for every single nominee, similar to what you did in `get_all_nominees_by_election`.

src/elections/urls.py
jbriones1: If you change `get_all_nominees_by_election` to use (left) `outerjoin`, then you can change this to `election_list = await (elections.crud.get_all_nominees_by_election(...) if has_permission else elections.crud.get_all_elections(...))`, and then shuffle the logic around a bit.

src/elections/urls.py
```diff
@@ -174,12 +139,14 @@ async def get_election(
...
     if has_permission:
         election_json = election.private_details(current_time)
-        if with_nominees:
-            election_json["candidates"] = await _get_election_nominees(db_session, election, has_permission)
     else:
         election_json = election.public_details(current_time)
-        if with_nominees:
-            election_json["candidates"] = await _get_election_nominees(db_session, election, has_permission)
+    if with_nominees:
+        nominees = await elections.crud._get_election_nominees(db_session, election, has_permission)
```
jbriones1: The underscore at the front of this function indicates it should only be used internally i.e. not outside of the function's module (file). Its not enforced by the interpreter so this runs fine, but you should remove the underscore so that its clear this is not an internal function.

src/nominees/urls.py
```diff
@@ -84,7 +84,7 @@ async def get_nominee_info(db_session: database.DBSession, computing_id: str):
     operation_id="delete_nominee",
     dependencies=[Depends(perm_election)],
 )
-async def delete_nominee_info(db_session: database.tDBSession, computing_id: str):
+async def delete_nominee_info(db_session: database.DBSession, computing_id: str):
```
jbriones1, line 87: Thanks for fixing that, I let that through >.<

jbriones1 (changes requested):
Thanks for changing things based on my feedback.

The tests look good. If you're not too busy and you're willing to, we can discuss how to populate the database with test data and we could make more comprehensive tests.

src/elections/crud.py
VerrillAng: Changed to `outerjoin`

src/elections/crud.py
VerrillAng: Changed to `outerjoin`

src/elections/crud.py
VerrillAng: Done

src/elections/urls.py
VerrillAng: Done

src/elections/urls.py
jbriones1: Just to clarify, you would also need to query the election information in `get_all_nominees_by_election`, so it might be worth it to make a new function that does all of that in one query.

src/elections/crud.py
```diff
@@ -37,3 +41,121 @@ async def delete_election(db_session: AsyncSession, slug: str) -> None:
     Deletes a given election by its slug. Does not validate if an election exists
     """
     await db_session.execute(sqlalchemy.delete(ElectionDB).where(ElectionDB.slug == slug))
+
+
+async def get_all_elections_with_nominees(
```
jbriones1, line 46: This function works, but I saw a less verbose way to handle this. I'll probably look into refactoring this when taking a look at the election tables again.

jbriones1 (approved):
LGTM, thank you!

PR #150, opened 2026-06-29
Add honourary member endpoints

### LINKED ISSUE #130: Create resources to support Honourary Members
### Database
* Create a migration with the table name `honorary_member` with config:
    * PrimaryKey(INTEGER): Term ID
    * Column(VARCHAR(128)): Name
    * Column(DATE): Start date
    * Column(DATE): End date
* Each time a membership is granted a new record must be created

### Endpoints
* Endpoint "/honourary"
* GET(/): get a list of all honourary members
* GET(/current): get a list of all active member terms
* POST(/): takes in a list of bodies, creates new member terms for each submission
* PATCH(/{term_id}): updates a term, restricted to terms that have not ended yet

AmrajKoonar:
Closes #130

Summary:
- Added an honorary_member table and Alembic migration
- Added honorary member models, CRUD functions, and routes
- Registered the /honorary router
- Added admin-only endpoints for listing, creating, updating, and deleting honorary member terms

Testing:
- Ran python -m pytest
- 47 tests passed


src/honourary/__init__.py
jbriones1, line 1: This file can be removed.

src/honourary/crud.py
```diff
@@ -0,0 +1,63 @@
...
+async def get_honorary_member_by_id_or_raise(
+    db_session: database.DBSession,
+    term_id: int,
+) -> HonoraryMemberDB:
+    honorary_member = await db_session.scalar(select(HonoraryMemberDB).where(HonoraryMemberDB.id == term_id))
+    if honorary_member is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"honorary member term with id={term_id} does not exist",
+        )
```
jbriones1: `<module>/crud.py` files shouldn't raise HTTP errors, this will be changed for the Officers and Nominees module in the future. Return `None` and have the error be raised in `honorary/urls.py`.

This is so unit tests for CRUD operation don't need to test `HTTPExceptions` that are meant for API endpoints.

src/honourary/crud.py
```diff
@@ -0,0 +1,63 @@
...
+    return honorary_member
+
+
+def ensure_term_has_not_ended(honorary_member: HonoraryMemberDB) -> None:
+    if honorary_member.end_date is not None and honorary_member.end_date < date.today():
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="cannot update or delete a term that has already ended",
+        )
```
jbriones1: Same reason as above You can change this to a function that returns a boolean and rename it to `has_term_ended(honorary_member: HonoraryMemberDB)` and then raise the `HTTPException` where needed.

Also, this should return status code 409 Conflict instead of 403 Forbidden. We use 409 to indicate that the request is well-formed, but the state of the resource doesn't allow it, while 403 Forbidden is when someone is authenticated, but not authorized to carry out an action.

src/honourary/models.py
```diff
@@ -0,0 +1,28 @@
+from datetime import date
+
+from pydantic import BaseModel, ConfigDict, Field, computed_field
+
+from utils import is_active_term
+
+
+class HonoraryMemberCreate(BaseModel):
+    name: str = Field(..., max_length=128)
```
jbriones1: Use `HONORARY_MEMBER_MAX_LENGTH = 128`.

src/alembic/versions/3c00f28c26e2_create_honorary_member_table.py
```diff
@@ -0,0 +1,36 @@
...
+down_revision: Union[str, None] = '42f855bec532'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table('honorary_member',
+    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
+    sa.Column('name', sa.String(length=128), nullable=False),
```
jbriones1: For the max length, make a file in the honorary folder and call it constants.py. Define a constant `HONORARY_MEMBER_MAX_LENGTH = 128` and then use it in the models file, table definition, and this file.

src/honourary/tables.py
```diff
@@ -0,0 +1,15 @@
...
+from sqlalchemy.orm import Mapped, mapped_column
+
+from database import Base
+
+
+class HonoraryMemberDB(Base):
+    __tablename__ = "honorary_member"
+
+    id: Mapped[int] = mapped_column(Integer, primary_key=True, autoincrement=True)
+    name: Mapped[str] = mapped_column(String(128), nullable=False)
```
jbriones1: Use `HONORARY_MEMBER_MAX_LENGTH = 128`.

src/honourary/crud.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+async def create_honorary_members(
+    db_session: database.DBSession,
+    honorary_members: list[HonoraryMemberDB],
+) -> list[HonoraryMemberDB]:
+    db_session.add_all(honorary_members)
+    await db_session.flush()
```
jbriones1: Remove the flush here and add it wherever it needs to happen.

src/honourary/urls.py
jbriones1, line 1: Add a `responses` dict to each endpoint to indicate what status codes can be returned. See https://github.com/CSSS/csss-site-backend/blob/aac1432df0ac0ef331faa87fd07fbda23dfbb987/src/nominees/urls.py#L112 for an example.

jbriones1 (changes requested):
Thanks for the PR, looks really good!

One thing that will need to change in many places is that "honourary" is less common than "honorary", [even in Canada](https://our-languages.canada.ca/en/writing-tips-plus/honor-honorary-honour-honourary). That's my bad since that's how I wrote it in the issue. Please go through and change everything that says "honourary" to "honorary"

AmrajKoonar:
Done

AmrajKoonar:
Fixed merge conflicts

src/alembic/versions/3c00f28c26e2_create_honorary_member_table.py
```diff
@@ -5,32 +5,36 @@
...
-    op.create_table('honorary_member',
-    sa.Column('id', sa.Integer(), autoincrement=True, nullable=False),
-    sa.Column('name', sa.String(length=128), nullable=False),
-    sa.Column('start_date', sa.Date(), nullable=False),
-    sa.Column('end_date', sa.Date(), nullable=True),
-    sa.PrimaryKeyConstraint('id', name=op.f('pk_honorary_member'))
+    op.create_table(
+        "honorary_member",
+        sa.Column("id", sa.Integer(), autoincrement=True, nullable=False),
+        sa.Column("name", sa.String(length=HONORARY_MEMBER_MAX_LENGTH), nullable=False),
```
jbriones1: I made a mistake, this should be hard-coded to 128. Keep the constant in the other files, though (sorry).

src/alembic/versions/3c00f28c26e2_create_honorary_member_table.py
```diff
@@ -5,32 +5,36 @@
...
+from alembic import op
+from honorary.constants import HONORARY_MEMBER_MAX_LENGTH
 
 # revision identifiers, used by Alembic.
-revision: str = '3c00f28c26e2'
-down_revision: Union[str, None] = '42f855bec532'
-branch_labels: Union[str, Sequence[str], None] = None
-depends_on: Union[str, Sequence[str], None] = None
+revision: str = "3c00f28c26e2"
+down_revision: str | None = "42f855bec532"
```
jbriones1: To fix the alembic CI failure you'll need to rebase this revision off the one on main.

If you run `alembic heads`  you'll see two different hashes. One of them is from the migration you created and the other is from the one currently on `main`. Take the revision of the head from main and replace the one here.

jbriones1 (changes requested):
Your patch looks great, but one project-specific things are missing.

 `src/load_test_db.py` needs to import your table file to ensure the test database creates it.

src/alembic/versions/3c00f28c26e2_create_honorary_member_table.py
AmrajKoonar: Done

src/alembic/versions/3c00f28c26e2_create_honorary_member_table.py
AmrajKoonar: Done

jbriones1 (approved):
Looks good, thanks!

PR #154, opened 2026-07-02
feat: Authentication, site user, and environment changes

### LINKED ISSUE #119: Improve our authentication
# TL;DR  We probably need to re-do our authentication.

Look at the [implementation section](#implementation-phases) for how we could implement it
https://fastapi.tiangolo.com/advanced/security/oauth2-scopes/
https://www.rfc-editor.org/rfc/rfc7519#section-10.1
https://www.rfc-editor.org/rfc/rfc8725

## Description
I think our backend should be frontend agnostic, which means we should probably improve how we handle sessions. We could also share sessions between different applications across the CSSS. This will require us to rewrite how our authentication flow works. Since we rely on SFU's CAS system, we would just need to adapt how our sessions are managed.

## Problems and Solution
There are a number of problems with our current implementation.

### 1. Strings are handled poorly
We need to check things are encoded properly and that the POST body from login is not malicious. 

#### Solutions
- [ ] (`login_user`) Remove `redirect_url`, just redirect to the `https://admin.sfucsss.org` for now,
- [ ] (`login_user`) `urllib.parse.quote(body.ticket)`  as well, just like `service`

### 2. Cleaning up expired sessions from the database is reliant on logins
If someone's session expires then it sits in the database until someone else logs in and the sessions are checked.

#### Solutions
- [ ] Add a CRON job that will sweep the table regularly

### 3. Site user pics are vulnerable to XSS
People could upload pictures that is actually malicious code.

#### Solution
- [ ] Remove site user pictures from the database and web server.

### 4. Logging in invalidates previous sessions
That means if someone has multiple tabs open they could be logged out of their session on certain tabs.

#### Solution

- [ ] Either we add Single Sign On (SSO) for all of our apps, but we only really care about the admin app, so just restrict the application to just that for now.

### 5. Session IDs are way too big
It's 256 bytes right now, which is apparently way too big.

#### Solution
- [ ] Change it to be 32 bytes
- [ ] Update the database table column to use the new size
- [ ] Update the Pydantic models to use the new size

### 6. No rate-limiting to our `login` endpoint
Our web server could be used to DoS the SFU CAS servers. This will probably need to be handled on Nginx.

#### Solution
- [ ] Add rate-limiting to Nginx and __maybe__ to the web server.

jbriones1:
Closes #119

## Description

Reworks authentication into a frontend-driven API flow, tightens session handling, removes stale site-user profile-picture support, and introduces typed environment-variable configuration. The branch also adds database migrations for the auth schema changes, integration coverage for the login/logout workflow, and updates CI to use `uv`.

## Summary of Changes

* Added Pydantic settings support for environment variables and local `.env` files.
* Added a root `.env.example` and updated README setup instructions.
* Replaced the old redirect-based login response with a `204 No Content` API response.
* Increased session token entropy and made session lifetime consistent across cookies and database checks.
* Removed site-user profile-picture support and the stale user update endpoint.
* Added Alembic migrations for session ID length and profile-picture removal.
* Added `auth` integration tests and a stricter pytest configuration.
* Updated GitHub Actions workflows to use `uv`.
* Removed dead auth and utility code.

## Environment Variables

* Added `src/config.py` settings for:
  * `ENVIRONMENT`
  * `DB_PORT`
  * `COOKIE_SECURE`
  * `COOKIE_DOMAIN`
  * `FRONTEND_ORIGIN`
  * `AUTH_URL`
  * `TRANSLINK_API_KEY`
* Added support for a root `.env` file.
* Added `.env.example` to document expected local settings.
* Replaced the previous `LOCAL=true` workflow with `ENVIRONMENT`:
  * `ENVIRONMENT=dev` runs the app in local/dev mode, using the `main` database.
  * `ENVIRONMENT=prod` disables OpenAPI docs.
  * `ENVIRONMENT=test` runs the app in local/dev mode, using the `test` database.
* Moved database URL selection to use configured `DB_PORT`.
* Configured CORS to accept requests only from `FRONTEND_ORIGIN`.
  * Currently only one frontend origin is supported.

## Authentication Flow

* Changed `POST /auth/login` to validate the CAS ticket, set the session cookie, and return `204 No Content`.
* Removed the forced backend redirect from `POST /auth/login`; frontend clients now handle navigation after login.
* Removed `redirect_url` from the login request body.
* Added explicit documented login responses for:
  * `204 No Content` on success
  * `401 Unauthorized` for CAS authentication failure
  * `502 Bad Gateway` for CAS connection, response, or parse failures
  * `503 Service Unavailable` when auth is not configured
* URL-encoded CAS `service` and `ticket` values before validating with CAS.
* Added defensive parsing for CAS response shape before reading nested fields.
* Avoided logging full CAS responses and ticket values.
* Updated auth status-code handling to use FastAPI's `status` constants.

## Sessions and Cookies

* Increased session IDs to 32 random bytes encoded as 43-character URL-safe base64 values.
* Updated `SESSION_ID_LEN` from `512` to `43`.
* Added a two-hour `Max-Age` to the session cookie.
* Added shared auth constants for the session cookie name and lifetime.
* Made database-backed session expiry use the same two-hour lifetime as the cookie.
* Fixed expired sessions still returning computing IDs or site-user records.
* Made `GET /auth/logout` idempotent.
* Fixed logout cookie deletion when `COOKIE_DOMAIN` is configured.
* Made missing or stale session deletion safe instead of raising on `None`.
* Added upsert-based site-user and session creation so repeated logins update existing records cleanly.

## Site User

* Removed site-user profile-picture support from the SQLAlchemy table and Pydantic response model.
* Removed the stale `PATCH /auth/user` endpoint.
* Removed the old `update_site_user` helper.
* Updated `SiteUser` and `UserSession` Pydantic models to validate ORM attributes.
* Allowed `first_logged_in` and `last_logged_in` to be nullable in the API model to match the database schema.
* Removed unused auth helper types.

## Database Migrations

* Added a migration to reduce `user_session.session_id` from 512 characters to the new 43-character session ID length.
* Added a migration to remove `site_user.profile_picture_url`.

## Tests

* Added integration tests for:
  * successful CAS login and cookie creation
  * fetching the current user from the session cookie
  * invalid CAS tickets
  * CAS upstream failures
  * missing auth configuration
  * idempotent logout and cookie deletion
  * expired session rejection
* Updated pytest configuration with:
  * strict markers
  * strict config
  * short tracebacks
  * explicit async test loop scope
  * documented `unit` and `integration` markers

## GitHub Actions

* Updated unit test, Ruff, and Alembic workflows to use `uv`.
* Added required environment variables to CI jobs so settings can be constructed during imports.
* Replaced manual Python/pip/venv setup in the Alembic workflow with `uv`.
* Replaced the standalone Ruff action with `uv run ruff check .`.

## Cleanup

* Removed stale imports and unused helpers.
* Removed outdated `LOCAL`/`IS_PROD` cookie and CORS constants.
* Updated README project structure documentation.


jbriones1:
To the reviewers, please either mark this as approved or leave questions if you're not sure about something.

PR #157, opened 2026-07-15
Moved Translink API under authenticated kiosk routes

kathryntanardy:
Close #155 

Move the TransLink API under a protected kiosk route using Bearer token authentication:
- Added kiosk Bearer auth using `KIOSK_SECRET`
- Created `/kiosk` router with auth applied globally
- Mounted TransLink routes at `/api/kiosk/translink/*`
- Removed public TransLink router mounting from `main.py`


jbriones1 (approved):
LGTM. When I fix my laptop I'll test this out.

PR #173, opened 2026-08-23
Include SAST and SCA

daeisbae:
This is a pull request to make sure the libraries and up to date so it's not vulnerable and codebase does not contain vulnerable coding pattern.


jbriones1 (changes requested):
I think we need to pin commit SHAs if we're using scanning tools, just to ensure we're not vulnerable to compromised new versions.

PR #177, opened 2026-08-27
Fix officers/all response for legacy positions

### LINKED ISSUE #159: Endpoint `/officers/all` on the production database returns 500
## Description
[https://sfucsss.org/api/officers/all](https://sfucsss.org/api/officers/all) returns 500.

Our REST API does a validation check that the values here are invalid, but old entries don't conform to those. It's probably best to remove the validation check and change the model to be a union of the enum types or a string.

## Error
```sh
Input should be 'president', 'vice-president', 'treasurer', 'director of resources', 'director of events', 'director of educational events', 'assistant director of events', 'director of communications', 'director of multimedia', 'director of archives', 'executive at large', 'first year representative', 'election officer', 'sfss council representative', 'frosh week chair', 'system administrator', 'webmaster' or 'social media manager' [type=enum, input_value='elections officer', input_type=str]
```

kathryntanardy:
Close #159 

Fix for the /officers/all 500 caused by legacy officer position values such as "elections officer". The Officer response model now treats position as a string, so old database values can be returned without being validated against the current OfficerPositionEnum. Write models still use the enum, so new create/update inputs remain strict.

### DIFF
--- a/src/officers/models.py
+++ b/src/officers/models.py
@@ -117,11 +117,7 @@ class Officer(BaseModel):
     @computed_field
     @property
     def csss_email(self) -> str | None:
-        try:
-            position = OfficerPositionEnum(self.position)
-        except ValueError:
-            return None
-        return OfficerPosition.to_email(position)
+        return OfficerPosition.to_email(self.position)
 
     term_id: int
 
--- a/src/officers/models.py
+++ b/src/officers/models.py
@@ -90,7 +90,14 @@ class OfficerBase(BaseModel):
     biography: str | None = None
 
 
-class Officer(OfficerBase):
+class Officer(BaseModel):
+    legal_name: str = Field(..., max_length=OFFICER_LEGAL_NAME_MAX)
+    position: str
+    start_date: date
+    end_date: date | None = None
+    nickname: str | None = None
+    biography: str | None = None
+
     @classmethod
     def public_fields(cls, term: OfficerTermDB, info: OfficerInfoDB) -> Self:
         return cls(
@@ -110,7 +117,11 @@ class Officer(OfficerBase):
     @computed_field
     @property
     def csss_email(self) -> str | None:
-        return OfficerPosition.to_email(self.position)
+        try:
+            position = OfficerPositionEnum(self.position)
+        except ValueError:
+            return None
+        return OfficerPosition.to_email(position)
 
     term_id: int
 
--- a/src/officers/constants.py
+++ b/src/officers/constants.py
@@ -46,7 +46,7 @@ class OfficerPosition:
             return _LENGTH_MAP[position]
 
     @staticmethod
-    def to_email(position: OfficerPositionEnum) -> str | None:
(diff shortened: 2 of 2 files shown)


src/officers/models.py
```diff
@@ -90,7 +90,14 @@ class OfficerBase(BaseModel):
     biography: str | None = None
 
 
-class Officer(OfficerBase):
+class Officer(BaseModel):
+    legal_name: str = Field(..., max_length=OFFICER_LEGAL_NAME_MAX)
+    position: str
```
karyao: I think `Officer` can inherit `OfficerBase` still just override the position field such as:

```
class Officer(OfficerBase):
     position: str

     // left unchanged
```

without having to redefine the existing fields

src/officers/models.py
```diff
@@ -90,7 +90,14 @@ class OfficerBase(BaseModel):
     biography: str | None = None
 
 
-class Officer(OfficerBase):
+class Officer(BaseModel):
```
jbriones1: You can keep this inheriting `OfficerBase` and just declare the `position` as a `str`.

src/officers/models.py
```diff
@@ -110,7 +117,11 @@ def is_active(self) -> bool:
     @computed_field
     @property
     def csss_email(self) -> str | None:
-        return OfficerPosition.to_email(self.position)
```
jbriones1: I think you can just change `OfficerPosition.to_email()` to accept a string instead of `OfficerPositionEnum`. It defaults to `None`, so there's no need to check if the StrEnum exists.

jbriones1 (changes requested):
I agree with your assessment that we should keep future inserts/updates using the positions we have defined.

src/officers/models.py
jbriones1: To ensure the other fields on this model don't drift, you should have this inherit from `OfficerBase` but then add `position: str`.

jbriones1 (changes requested):

src/officers/models.py
```diff
@@ -91,6 +91,8 @@ class OfficerBase(BaseModel):
 
 
 class Officer(OfficerBase):
+    position: str  # pyright: ignore[reportIncompatibleVariableOverride]
```
jbriones1: Hm, I see that this causes an issue.

We might as well just make OfficerBase's position `str` for now. I'll go through the database at some point to normalize all the positions and then we can tighten up the types then.

src/officers/models.py
jbriones1:
```suggestion
    position: str
```
See below for reasoning

jbriones1 (changes requested):
Sorry, I keep waffling on this. Last one, I swear >.<

jbriones1:
@kathryntanardy I think this issue exposes a deeply rooted issue that we have a chance to address now since we'll be migrating our infrastructure. This means I can re-do our schema and prepare the data so this can be addressed properly. I'm going to pause this issue for now

PR #178, opened 2026-08-27
Event info table update

### LINKED ISSUE #156: Event database table update
## Description

We currently have no way of assigning a poster to an event. The events will live on this web server for now, but may change later (no need to worry about this).

### Database schemas

See https://github.com/CSSS/csss-site-backend/issues/156#issuecomment-5305409648

### Event Endpoints
I think most of the endpoints are fine for `event_info`, but **GET /events** should also include the `image_url` of the event.

I would add some way to perform CRUD operations on a group of events.

1. **POST /event/group**
Creates one or more events under the same group key.
Body: Array of `EventCreate`

Response:
```ts
{
  "group_id": string;
  "events": Event[]
}
```
2. **PATCH /event/group/{group_id}**
Updates one more more events under the same group.
```js
Event[]
```

3. **DELETE /event/group/{group_id}**
Deletes one more more events under the same group. Responses should follow the same patterns as before.

arunPdl02:
<html>
<body>
Fixes #156
<html><head></head><body><h1>Summary</h1>
<p>Adds three new endpoints for managing grouped events, along with a schema migration to <code>event_info</code> that replaces recurrence-based fields (<code>frequency</code>, <code>repeat_start_date</code>, <code>repeat_end_date</code>, <code>start_time</code>, <code>end_time</code>) with a group-based model (<code>group_id</code>) plus new metadata fields (<code>location</code>, <code>organizer</code>, <code>status</code>, <code>url</code>, <code>image_id</code>). Also refactors <code>crud.py</code> for readability.</p>
<hr>
<h2>New Endpoints</h2>
<h3><code>POST /event/group</code></h3>
<p>Creates one or more events under the same group key.</p>
<p><strong>Request body:</strong> <code>EventCreate[]</code></p>
<p><strong>Response:</strong></p>
<pre><code class="language-json">{
  "group_id": "string",
  "events": [ /* Event[] */ ]
}
</code></pre>
<h3><code>PATCH /event/group/{group_id}</code></h3>
<p>Updates one or more events belonging to the given group.</p>
<p><strong>Response:</strong></p>
<pre><code class="language-json">{
  "group_id": "string",
  "events": [ /* Event[] */ ]
}
</code></pre>
<h3><code>DELETE /event/group/{group_id}</code></h3>
<p>Deletes one or more events belonging to the given group. Response follows the same shape as the endpoints above.</p>
<p><strong>Response:</strong></p>
<pre><code class="language-json">{
  "group_id": "string",
  "events": [ /* Event[] */ ]
}
</code></pre>
<hr>
<h2>Database Changes</h2>
<h3><code>event_info</code> schema</h3>

Column | Type | Nullable | Notes
-- | -- | -- | --
eid | int | No | Autoincrement, PK
description | text | No |  
name | text | No | Event title (widened from varchar(64) → text)
start_datetime | timestamptz | No | Replaces start_time/repeat_start_date
end_datetime | timestamptz | No | Replaces end_time/repeat_end_date
group_id | uuid | Yes | Links events that should update/delete together
location | text | Yes | Room number or full address
organizer | text | Yes | e.g. CSSS, SFU, Faculty of Applied Sciences, a club
status | enum (scheduled, cancelled) | No |  
url | text | Yes | Link to event resource (site, signup form, etc.)
image_id | int (FK → image_asset.image_id) | Yes | Event poster

<h3>Migration (<code>41152814ded4</code>)</h3>
<p><strong>Added:</strong></p>
<ul>
<li><code>start_datetime</code> (timestamptz, not null)</li>
<li><code>end_datetime</code> (timestamptz, not null)</li>
<li><code>group_id</code> (uuid, nullable)</li>
<li><code>location</code> (text, nullable)</li>
<li><code>organizer</code> (text, nullable)</li>
<li><code>status</code> (enum: <code>scheduled</code> / <code>cancelled</code>, not null)</li>
<li><code>url</code> (text, nullable)</li>
<li><code>image_id</code> (int, nullable) + FK constraint to <code>image_asset.image_id</code></li>
</ul>
<p><strong>Altered:</strong></p>
<ul>
<li><code>description</code>: now <code>NOT NULL</code></li>
<li><code>name</code>: <code>varchar(64)</code> → <code>text</code></li>
</ul>
<p><strong>Removed:</strong></p>
<ul>
<li><code>repeat_end_date</code></li>
<li><code>start_time</code></li>
<li><code>end_time</code></li>
<li><code>frequency</code></li>
<li><code>repeat_start_date</code></li>
</ul>
<hr>
<h2>Refactor</h2>
<p>Broke down <code>crud.py</code> functions that previously did query + result handling + event formatting in a single line into separate, named steps (query → result → event format) for readability.</p></body></html>
</body>
</html>


src/event/constants.py
```diff
@@ -8,3 +8,7 @@ class EventFrequencyEnum(StrEnum):
     MONTHLY = "MONTHLY"
     SEMESTERLY = "SEMESTERLY"
     YEARLY = "YEARLY"
```
jbriones1: You can remove these.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
+    db_session: AsyncSession,
+    eid: int
+) -> EventDB | None:
+    query = select(EventDB).where(EventDB.eid == eid)
+
+    result = await db_session.execute(query)
+
+    event = result.scalar_one_or_none()
+
+    return event
```
jbriones1:
```suggestion
return await db_session.get(EventDB, eid)
```

This is more modern syntax that works for simple cases. Eventually the rest of the repo should query this way when using primary keys.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
+    db_session: AsyncSession,
+    group_id: UUID
+) -> Sequence[EventDB] | None:
+    query = select(EventDB).where(EventDB.group_id == group_id)
+
+    result = await db_session.execute(query)
+
+    event = result.scalars().all()
+
+    return event
```
jbriones1:
```suggestion

    return result.scalars().all()
```

Minor nitpick, reduces the amount of assignments used by 1, but it's probably already handled by the interpreter.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
 async def create_event(db_session: AsyncSession, info: EventDB):
     db_session.add(info)
 
 
-async def delete_event(db_session: AsyncSession, eid: int):
-    result = await db_session.execute(delete(EventDB).where(EventDB.eid == eid))
+async def delete_event(
+    db_session: AsyncSession,
+    eid: int
+):
```
jbriones1: Missing a return type.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
+
+    result = await db_session.execute(query)
+    # Return the number of rows affected
+    return result.rowcount
+
+
+async def delete_group_events(
+    db_session: AsyncSession,
+    group_id: UUID
+):
```
jbriones1: Missing a return type.

src/event/models.py
```diff
@@ -1,31 +1,26 @@
...
+    group_id: UUID | None = None
+    location: str | None = None
+    organizer: str | None = None
+    status: EventStatusEnum
+    url: str | None = None
 
     @model_validator(mode="after")
     def validate_time_range(self) -> "BaseEvent":
-        if self.start_time >= self.end_time:
+        if self.start_datetime >= self.end_datetime:
```
jbriones1:
```suggestion
        if self.start_datetime > self.end_datetime:
```

Based on the schema and spec, `start_datetime` and `end_datetime` being equal is valid.

src/event/models.py
```diff
@@ -37,18 +32,37 @@ class Event(BaseEvent):
 class EventCreate(BaseEvent):
     pass
 
+class GroupEventCreate(BaseModel):
```
jbriones1:
```suggestion
class GroupEvent(BaseModel):
```

This is a response, usually we've used `TableCreate` to mean "these are the params the client needs to send us to create the objects.

src/event/models.py
```diff
@@ -1,31 +1,26 @@
...
-    repeat_start_date: datetime.date | None = None
-    repeat_end_date: datetime.date | None = None
+    description: str
+    start_datetime: datetime.datetime
+    end_datetime: datetime.datetime
+    group_id: UUID | None = None
+    location: str | None = None
+    organizer: str | None = None
+    status: EventStatusEnum
+    url: str | None = None
```
jbriones1: Missing `image_id`.

src/event/models.py
```diff
@@ -37,18 +32,37 @@ class Event(BaseEvent):
...
+    status: EventStatusEnum | None = None
+    url: str | None = None
 
 
 class EventDelete(BaseModel):
     result: bool
     eid: int
+
+
+class GroupEventDelete(BaseModel):
```
jbriones1:
```suggestion
class GroupEventDeleteResponse(BaseModel):
```

Similar to `GroupEventCreate`, just so we can differentiate between them.

src/event/models.py
```diff
@@ -37,18 +32,37 @@ class Event(BaseEvent):
...
+    result: bool
+    group_id: UUID
+    event_deleted: int
```
jbriones1:
```suggestion
    events_deleted: int[]
```
This should be an array so you can return the `eid` of all events deleted.

src/event/urls.py
```diff
@@ -129,3 +209,21 @@ async def delete_event(db_session: database.DBSession, eid: int):
...
+    dependencies=[Depends(perm_admin)],
+)
+async def delete_group_event(db_session: database.DBSession, group_id: uuid.UUID):
+    rows_deleted = await event.crud.delete_group_events(db_session, group_id)
+
+    if rows_deleted == 0:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Event doesn't exist.")
+
+    await db_session.commit()
+    return GroupEventDelete(result=True, group_id=group_id, event_deleted= rows_deleted)
```
jbriones1: Return the IDs deleted, so clients know what was successfully removed.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
-    result = await db_session.execute(delete(EventDB).where(EventDB.eid == eid))
+async def delete_event(
+    db_session: AsyncSession,
+    eid: int
+):
+    query = delete(EventDB).where(EventDB.eid == eid)
+
+    result = await db_session.execute(query)
+    # Return the number of rows affected
+    return result.rowcount
```
jbriones1: Instead of returning the row count return the `eid`.

See `src/auth/crud.py: 144` to see how you can extract the `eid` when deleting.

src/event/crud.py
```diff
@@ -31,28 +41,69 @@ async def get_events_for_this_year_month(
...
+
+async def delete_group_events(
+    db_session: AsyncSession,
+    group_id: UUID
+):
+    query = delete(EventDB).where(EventDB.group_id == group_id)
+
+    result = await db_session.execute(query)
     # Return the number of rows affected
     return result.rowcount
```
jbriones1: Instead of returning the row count return the `eid` of all events deleted.

See `src/auth/crud.py: 144` to see how you can extract the `eid` when deleting.

src/event/tables.py
```diff
@@ -1,26 +1,41 @@
...
-    start_time: Mapped[datetime] = mapped_column(DateTime(timezone=True))
-    end_time: Mapped[datetime] = mapped_column(DateTime(timezone=True))
-    frequency: Mapped[EventFrequencyEnum] = mapped_column(String(64), server_default=text("'NONE'"))
-    repeat_start_date: Mapped[date] = mapped_column(Date, nullable=True)
-    repeat_end_date: Mapped[date] = mapped_column(Date, nullable=True)
+    description: Mapped[str] = mapped_column(Text)
+    name: Mapped[str] = mapped_column(Text)
+    start_datetime: Mapped[datetime] = mapped_column(DateTime(timezone=True))
+    end_datetime: Mapped[datetime] = mapped_column(DateTime(timezone=True))
+    group_id: Mapped[UUID] = mapped_column(Uuid, nullable=True)
```
jbriones1:
```suggestion
    group_id: Mapped[UUID] = mapped_column(Uuid, nullable=True, index=True)
```

Add an index on this field.

src/event/urls.py
```diff
@@ -113,6 +147,52 @@ async def update_event(db_session: database.DBSession, eid: int, body: EventUpda
...
+
+    db_events = await event.crud.get_events_by_group_id(db_session, group_id)
+    if not db_events:
+        raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Events with this group_id don't exist.")
+
+    patch_data = body.model_dump(exclude_unset=True)
+
+    validated_events = []
+    for ev in db_events:
+        db_data = Event.model_validate(ev).model_dump()
```
jbriones1: Don't think we need to validate when it comes from the database.

src/event/urls.py
```diff
@@ -80,6 +82,38 @@ async def create_event(db_session: database.DBSession, body: EventCreate):
...
+    dependencies=[Depends(perm_admin)],
+)
+async def create_group_events(db_session: database.DBSession, body: list[EventCreate]):
+    new_event_list = []
+    g_id = uuid.uuid4()
+    for e in body:
+        e.group_id = g_id
+        new_event = EventDB(**e.model_dump())
+
+        await event.crud.create_event(
```
jbriones1: Instead of repeatedly calling `db_session.add(...)`, you should do a bulk insert using `add_all(...)` once you've constructed all of the EventDB objects.

You could also change this for loop to use list comprehension to speed it up more. Probably doesn't matter for our small database, but why not.

src/event/urls.py
```diff
@@ -113,6 +147,52 @@ async def update_event(db_session: database.DBSession, eid: int, body: EventUpda
...
+    validated_events = []
+    for ev in db_events:
+        db_data = Event.model_validate(ev).model_dump()
+        merged_data = {**db_data, **patch_data}
+        try:
+            Event.model_validate(merged_data)
+        except ValidationError as e:
+            raise HTTPException(
+                status_code=status.HTTP_422_UNPROCESSABLE_ENTITY, detail=jsonable_encoder(e.errors())
+            ) from e
```
jbriones1: I think FastAPI already handles errors thrown from `model_validate()`.

src/event/urls.py
```diff
@@ -113,6 +147,52 @@ async def update_event(db_session: database.DBSession, eid: int, body: EventUpda
...
 
+@router.patch(
+    "/group/{group_id}",
+    description="Update Event detail for all events with the given group_id",
+    response_model=GroupEventCreate,
+    responses={404: {"description": "Event doesn't exist."}},
+    operation_id="update_group_events",
+    dependencies=[Depends(perm_admin)],
+)
+async def update_group_events(
```
jbriones1: I think this function should take a list of `EventUpdates` and update the corresponding events.

For example, I have a group A, with events 1, 3, and 5.
I send PATCH `/group/A` with patched data for events 1 and 5, then only 1 and 5 would be updated. The response would contain a list of all the updated bodies, no need to send back the group ID as a separate field.

I think if the body contained events 1, 2, and 5 then we would reject it with 400, so there should be a check that the group we are targeting and the bodies we sent up match.

The response also contains the group ID, but the patch bodies sent up can change the group IDs of the events. I think we may need to make groups IDs not editable on this endpoint, so the patch model shouldn't contain group ID.

I've updated the issue to reflect what the return body should be.

jbriones1 (changes requested):
Thanks for the PR!

I think we'll need to a bit more work on the table to make it more ergonomic, but it's a good first pass. I'll make another issue with more endpoints we could use, but don't worry about it for now.

src/event/urls.py
arunPdl02: I tried and the server throws 500 instead of 422. FASTAPI didn't catch it.

jbriones1 (approved):

